PR #343, opened 2012-11-28
SPDY Support

kennydude:
I think I've got SPDY working on the https proxy.

It uses node-spdy to do the actual spdy stuff, but when testing with the https-to-https example Chrome flags SPDY is working


spillevink:
SPDY/HTTP2 is becoming more widespread, is it still out of scope for this project? I'd greatly appreciate if you'd reconsider.

PR #572, opened 2014-02-03
Issue #570: Use http.globalAgent as agent if no agent is explicitly specified.

### LINKED ISSUE #570: node-http-proxy leaks file descriptors
Run this:

```
http = require('http');
httpProxy = require('http-proxy');

proxy = httpProxy.createProxyServer();
server = http.createServer(function(req, res) {
    return proxy.web(req, res, {
      target: 'http://example.com'
    });
});

server.listen(9000);
```
- `ps -ef | grep node` to figure out the PID that's running it.
- `watch 'lsof -p 5966 | wc -l'` (replace 5966 with your pid)
- Point a browser at http://localhost:9000 (you'll get a 404 from example.com).  I was using Chrome and FF.
- Hit CTRL-R over and over.
- Watch your open file count climb.

Sockets don't seem to close for 2 minutes.  On OS/X, heavy traffic will run out of file descriptors very quickly (my product's Selenium test case kills node-http-proxy about 1/10th of the way through.)

Using node v0.10.25 and node-http-proxy 1.0.2.

jwalton:
Make it so we default to [`http.globalAgent`](http://nodejs.org/api/http.html#http_http_globalagent) as the agent if no agent is explicitly specified.  This is the agent that HTTP requests use by default when no agent is specified, so this is more in keeping with the http module, and it also fixes #570.  :)

`globalAgent` was introduced way back in [node 0.5.3](http://nodejs.org/docs/v0.5.3/api/http.html#http.globalAgent), so this should be a reasonably safe change as far as compatibility goes.

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -1,5 +1,6 @@
 var common = exports,
     url    = require('url'),
+    http   = require('http'),
     extend = require('util')._extend;
 
 /**
@@ -42,8 +43,8 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
     outgoing.rejectUnauthorized = (typeof options.secure === "undefined") ? true : options.secure;
   }
 
+  outgoing.agent = (options.agent == null) ? http.globalAgent : options.agent
 
-  outgoing.agent = options.agent || false;
   outgoing.path = url.parse(req.url).path;
   return outgoing;
 };


szubster:
Could using `globalAgent` be revisited?
the socket limit of 5 was changed to inifinity years ago
https://github.com/nodejs/node/commit/9fc9b87472806147b83c24d85b303c4f75d3021c#diff-82c531677b9ab4e260025e0397fb0ac4d4316459fa5d16346324e39584cfce96L73

PR #582, opened 2014-02-11
Add option for attaching outgoing headers, extraHeaders.

cheddar:
This also required adjusting web-incoming to pass the options object to the web-outgoing passes.  One thing I noticed is that web-outgoing doesn't appear to be exposed to anyone that wants mutate it externally.

I think extensions of this nature would be better done via some well-defined API that allows me to mutate the passes on the proxy object.  But, as it stands that doesn't exist and is too much for me to just guess at what would be acceptable and try to fix.


jcrugzz:
Also as a note, headers can be modified in the `proxyRes` event

PR #620, opened 2014-04-08
Gzip incoming

cliftonc:
We have a need to support a backend service that is returning responses that are already gzip encoded.  

This PR enables the incoming handler to recognise that the content is in gzip format and unzip it so that it can pass through subsequent middleware as normal.  If backend service does not return gzip encoded header then no change is made to the stream.

Minimal change to the incoming request handler, have added a test case to demonstrate it works (though I do get failing tests that I am unsure to fix that appears to mirror failing tests in the master branch).

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -60,6 +60,15 @@ var passes = exports;
   },
 
   /**
+   * Ensure content encoding is always plain on output 
+   * if input is gzip.
+   */
+  function gzipToPlain(req, res, proxyRes) {
+    if(proxyRes.headers["content-encoding"] == 'gzip') {
+      res.setHeader('content-encoding', 'text/plain');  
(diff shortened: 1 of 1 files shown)


pdf:
@cliftonc I'm confused as to why this was closed?

cliftonc:
I can't recall the specific reason, but I do remember that although this PR appeared to work I couldn't actually get the tests to pass and we encountered other issues.  In the end we just ended up not using http-proxy in the project, we just went down a more direct route:  https://github.com/tes/reliable-get.

pdf:
Hrm, according to travis it looks like the tests passed fine on node 0.10, they only failed on 0.11 due to some (unrelated?) HTTPS issue.  I don't really deal with node stuff generally, really just wanted this to work for grunt-connect to proxy some stuff through to my API server, but perhaps I'll try and cherry-pick your commits to re-send this PR.

PR #770, opened 2015-02-05
Add support for auto host rewriting and protocol rewriting

matthauck:
auto host rewriting allows rewriting to work as expected in most
cases without extra cumbersome configuration

protocol rewriting allows node-http-proxy to be able to listen
over HTTPS and properly reverse-proxy to sites running over HTTP
(to avoid doing SSL twice)


matthauck:
cancelling. going to resubmit from other account.

PR #774, opened 2015-02-12
Proxy for any target

gagern:
It would be nice if one could easily use the module to proxy not just for one target, but for any target the incoming connection requests. I.e. do normal web cache behaviour, not reverse proxying. Here is a setup to do so. Since several examples in the README use `{}` as the configuration, i.e. without `target` or `forward` setting, these examples will currently fail. With my addition, they will work and simply forward the requests.

Perhaps you might want to make people aware that proxying without a fixed target and without a bind restriction to localhost, any incoming request might query pages from the local LAN or even localhost. Depending on configuration, that may or may not be a security issue. If you consider this an issue, write examples in such a way that copying them will work (or fail due to an invented domain name) but will not cause such security problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,7 +340,10 @@ proxyServer.listen(8015);
  *  **secure**: true/false, if you want to verify the SSL Certs
  *  **xfwd**: true/false, adds x-forward headers
  *  **toProxy**: passes the absolute URL as the `path` (useful for proxying to proxies)
- *  **hostRewrite**: rewrites the location hostname on (301/302/307/308) redirects.
+ *  **prependPath**: specify whether you want to prepend the target's path to the proxy path
+ *  **localAddress**: Local interface string to bind for outgoing connections
+ *  **changeOrigin**: changes the origin of the host header to the target URL
+ *  **hostRewrite**: rewrites the location hostname on (301/302/307/308) redirects
 
 If you are using the `proxyServer.listen` method, the following options are also applicable:
 
--- a/lib/http-proxy.js
+++ b/lib/http-proxy.js
@@ -45,8 +45,8 @@ module.exports.createProxyServer =
    *  }
    *
    *  NOTE: `options.ws` and `options.ssl` are optional.
-   *    `options.target and `options.forward` cannot be
-   *    both missing
+   *    If neither `options.target` nor `options.forward` is given,
+   *    then any target server may be proxies.
    *  }
    */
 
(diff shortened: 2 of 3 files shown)


jcrugzz:
I'm not sure I understand the use case. If you are not proxying to a particular target, why the extra request overhead? Could you elaborate? Also please point out any examples that do not function :). All of them seem to specify a target in the constructor (for only proxying to one place) or in `proxy.web` where you can specify a target only using 1 instance.

gagern:
My use case was the following: I had some JavaScript deployed at a client's site, and wanted to try a newer development version in that environment. So I configured my browser to use a local proxy, and wrote that proxy such that it forwards all requests *except* for those matching a specific regular expression, which were served from a local file instead. The problem is that the moment I configure my browser to use that process as a proxy, *all* my web traffic goes through it, which is annoying if I want to look up some documentation while experimenting with the modified environment.

The way I did this was using `proxy.web(req, res, {target:req.url, prependPath:false})`. So yes, if you take `proxy.web` arguments into account, it seems that the `README.md` is more correct than it appeared to me. Sorry for the false alarm there. I guess I mixed some parts which were not intended to be mixed. If you don't want to merge my branch, then perhaps you should at least document the fact that a line like the one I mentioned will make the proxy behave like a web cache, forwarding any request it receives. It took me quite a while to work that out.

chenzx:
This is what situation i face: i'm a browser kernel developer from Lenovo, sometimes a specific page renders broken, we need to investigate the reason: whether it is some js file execution fails?

So i need to deploy a local http proxy server scripts, and forward all the requests to external web sites, except the hesitated js file request, which i will use a local served modified version.

I found https://github.com/substack/bouncy, but it is only used for load-balancing, hope http-proxy can support this feature, currently it seems not being able to deal with chunked-encoding?

prust:
> If you don't want to merge my branch, then perhaps you should at least document the fact that a line like the one I mentioned will make the proxy behave like a web cache

+1 for closing this PR and making it clearer in the docs that `target` is optional in the `createProxyServer()` options and that it can be passed in to `proxy.web()`. Like @gagern, I thought `target` was required; it took me a while to discover it could be generated dynamically and passed to `proxy.web()` for each request.

jcrugzz:
@prust Would happily accept a PR with updated Docs :+1:

indexzero:
Closing as per prior conversation. Feel free to open a follow-up PR. Thanks all!

PR #787, opened 2015-03-09
Fixes / additions to URL rewriting

matthauck:
* Add support for auto host rewriting and protocol rewriting
* Only rewrite URLs when the redirected host matches the original target


jcrugzz:
LGTM.

matthauck:
sweet, thanks!

PR #789, opened 2015-03-11
fix "x-forwarded-proto" in node 0.12 and iojs

### LINKED ISSUE #772: x-forwarded-proto incorrectly set to 'http' on node 0.12.0
I upgraded node to 0.12.0 and it looks like the https detection logic isn't working any more.

It looks like the issue is on line 70 of /lib/http-proxy/passes/web-incoming.js: req.connection.pair is evaluating to undefined even though it's an https connection.

I'm creating a server like so:

    https_server = https.createServer my_https_options, (req, res) ->
        proxy.web req, res, {
            xfwd: true
            target: {host: 'localhost', port: configuration.port}
        }

    https_server.listen configuration.ssl_port

feross:
The way to detect TLSSockets in node 0.12 and iojs has changed. You can
just check `socket.encrypted` now :)

Fixes #772

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -64,10 +64,11 @@ web_o = Object.keys(web_o).map(function(pass) {
   function XHeaders(req, res, options) {
     if(!options.xfwd) return;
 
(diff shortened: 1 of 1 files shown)


jcrugzz:
@feross awesome thanks! :)

PR #792, opened 2015-03-12
Adding the nodejs0.12 auth option

ashubham:
On the basis of https://nodejs.org/api/https.html#https_https_request_options_callback


lib/http-proxy/common.js
```diff
@@ -46,6 +46,10 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
     extend(outgoing.headers, options.headers);
   }
 
+  if(options.auth){
```
3rd-Eden: Please follow the code style in this document (adding spaces around the parentheses)

test/lib-http-proxy-common-test.js
```diff
@@ -17,6 +17,7 @@ describe('lib/http-proxy/common.js', function () {
         },
         headers: {'fizz': 'bang', 'overwritten':true},
         localAddress: 'local.address',
+	auth:'username:pass'
```
3rd-Eden: Use spaces instead of tabs for indenting here.

3rd-Eden:
It could use a test to ensure that the Authorization header is actually set instead of checking if the options are supplied.

jcrugzz:
LGTM

PR #798, opened 2015-04-01
Fix #747

### LINKED ISSUE #747: Canot call method 'split' of null
After recent upgrade my tests break:
```
Uncaught TypeError: Cannot call method 'split' of null
    at Object.common.urlJoin (/code/http-master/node_modules/http-proxy/lib/http-proxy/common.js:142:23)
    at Object.common.setupOutgoing (/code/http-master/node_modules/http-proxy/lib/http-proxy/common.js:75:26)
    at Array.stream [as 3] (/code/http-master/node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js:109:14)
    at ProxyServer.<anonymous> (/code/http-master/node_modules/http-proxy/lib/http-proxy/index.js:80:21)
    at Object.ProxyMiddleware.requestHandler (/code/http-master/modules/middleware/proxy.js:9:4607)
```

damonmcminn:
https://github.com/nodejitsu/node-http-proxy/issues/747

If `req.url` is an empty string, `url.parse(req.url).path`  is `null` and `common.urlJoin` will raise a `TypeError`.

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -80,7 +80,7 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
   // Remark: Can we somehow not use url.parse as a perf optimization?
   //
   var outgoingPath = !options.toProxy
-    ? url.parse(req.url).path
+    ? (url.parse(req.url).path || '/')
     : req.url;
 
   outgoing.path = common.urlJoin(targetPath, outgoingPath);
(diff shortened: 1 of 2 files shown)


jcrugzz:
@damonmcminn whats the use case of making req.url an empty string out of curiosity? Thanks for the fix!

jcrugzz:
fix in `1.9.1`

damonmcminn:
@jcrugzz No problem! I don't believe there is a use case for an empty string... It was an accidental discovery when I changed a regex that is used to strip the first path from `req.url`:
https://github.com/damonmcminn/api-proxy/commit/7dda2413535c5d5c1be0f574dde8ee69cd827fa7
The stripped path is used to identify what server to proxy the request to (so I can serve multiple APIs off a single validated domain i.e. api.damonmcminn.com) e.g.:
`api.damonmcminn.com/nutrition/green-turtle => localhost:50000/green-turtle`

PR #799, opened 2015-04-02
Fix default port detection with node 0.12.x

Volune:
With node 0.12.2, `common.getPort` was always returning a default port `80` even for HTTPS connections, because `req.connection.pair` is undefined on this version of node.

Here is a fix, along with some fixes for the HTTPS tests. It has been tested with node 0.10.38 and node 0.12.2 on windows.


jcrugzz:
@Volune this LGTM. Thanks for the thorough fix :+1:

PR #815, opened 2015-04-28
add way to modify request payload before send to target/forward; added s...

simone-sanfratello:
...tart event for this


jcrugzz:
Sorry this is a huge breaking change that I don't agree with. I would like to find a way to support this use case outside of http-proxy. There should be other modules to handle this async dance to make it work. I appreciate the contribution but will be closing this as I don't believe this is the right direction. For future contributions please try and follow the style in the project :). Thanks!

simone-sanfratello:
Thank you for your answer, I understand your point of view

PR #823, opened 2015-05-13
Added installation instructions

montogeek:
Installation process may be something obvious but is better if are documented.

I am creating this PR because I tried to install it using the package name `node-http-proxy` as the readme stands, but it happens to be the wrong name.

The correct name is `http-proxy`.


jcrugzz:
thanks! :smile:

montogeek:
You are welcome!

PR #825, opened 2015-05-18
Created reverse-proxy.js example.

pose:
Adds example of how to create a reverse proxy.


TKMacmillan:
Is it possible for this to work from the browser, not by issuing a curl request from the command line?

indexzero:
@TKMacmillan no since it relies on node libraries, not `XMLHttpRequest`.

TKMacmillan:
Without XMLHtttpRequest, if I were to enter localhost:8213 in my browser, could this script send the request elsewhere, like a web service at another location, and return the response?

PR #833, opened 2015-05-29
Don't overwrite global options

efokschaner:
Global options are being overwritten with every request causing per-request options to leak into the next call.

### DIFF
--- a/lib/http-proxy/index.js
+++ b/lib/http-proxy/index.js
@@ -46,9 +46,9 @@ function createRightProxy(type) {
         args[cntr] !== res
       ) {
         //Copy global options
-        options = extend({}, options);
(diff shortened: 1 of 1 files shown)


jcrugzz:
@efokschaner thanks for this! Could you provide a test case so we can defend against this?

efokschaner:
@jcrugzz Sorry for taking a while to get back to you. I cannot commit to adding tests for this as it is low priority for me and a non-trivial investment of my time given my lack of familiarity with node and javascript.

If I can, I will look at submitting some tests at a later date, but for now, please consider this PR for merge as is.

efokschaner:
Closing in light of #1074.

PR #836, opened 2015-06-02
Removed host proxy constraint

whatknight:
- Removed constraint that prevented rewrites when the target and redirected host did not match.
- Removed associated tests.

This constraint was added between v 1.9.x and 1.11.x and broke an implementation that was being used to proxy localhost requests to a another server.

If there's a good reason to keep preventing proxying between hosts that are not the same, I'd be interested to understand why and if there's a workaround for my specific use case.

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -50,13 +50,15 @@ var redirectRegex = /^30(1|2|7|8)$/;
     if ((options.hostRewrite || options.autoRewrite || options.protocolRewrite)
         && proxyRes.headers['location']
         && redirectRegex.test(proxyRes.statusCode)) {
-     
+      var target = url.parse(options.target);
       var u = url.parse(proxyRes.headers['location']);
 
       if (options.hostRewrite) {
         u.host = options.hostRewrite;
       } else if (options.autoRewrite) {
         u.host = req.headers['host'];
+      } else if (target.host != u.host) {
+        u.host = target.host;
       }
       if (options.protocolRewrite) {
         u.protocol = options.protocolRewrite;
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -50,13 +50,8 @@ var redirectRegex = /^30(1|2|7|8)$/;
     if ((options.hostRewrite || options.autoRewrite || options.protocolRewrite)
         && proxyRes.headers['location']
         && redirectRegex.test(proxyRes.statusCode)) {
-      var target = url.parse(options.target);
+     
       var u = url.parse(proxyRes.headers['location']);
-
-      // make sure the redirected host matches the target host before rewriting
-      if (target.host != u.host) {
-        return;
-      }
 
       if (options.hostRewrite) {
         u.host = options.hostRewrite;
--- a/test/lib-http-proxy-passes-web-outgoing-test.js
+++ b/test/lib-http-proxy-passes-web-outgoing-test.js
@@ -48,21 +48,7 @@ describe('lib/http-proxy/passes/web-outgoing.js', function () {
         this.options.autoRewrite = true;
         httpProxy.setRedirectHostRewrite(this.req, {}, this.proxyRes, this.options);
         expect(this.proxyRes.headers.location).to.eql('http://ext-manual.com/');
-      });
-
-      it('not when the redirected location does not match target host', function() {
-        this.proxyRes.statusCode = 302;
-        this.proxyRes.headers.location = "http://some-other/";
-        httpProxy.setRedirectHostRewrite(this.req, {}, this.proxyRes, this.options);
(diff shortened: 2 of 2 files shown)


jcrugzz:
@matthauck could you provide clarification for the commit here https://github.com/nodejitsu/node-http-proxy/commit/26029ba7ac948b5dc0befb2091cc9a5862d0641c? I personally don't have a strong opinion either way I just don't want to break it again and would like to find a way for this to work for both your cases

matthauck:
Ah yes. That was actually important. The justification was in the commit you mentioned. If the website is sending back a redirect to an external server, that should not be rewritten by the reverse proxy... This is similar to what apache does with ProxyPassReverse.

whatknight:
I would say that if you're explicitly setting a host rewrite, it's inferred that you want to rewrite the host with the value provided. 

Apache will rewrite incoming requests to make sure redirects always have the same host as `target`, so how about if we fall through the configurable options, we can set `u.host` to `target.host` if they don't match?

matthauck:
Hmm. My personal opinion is that it only makes sense to do the rewrite if it is a redirect to the same backend host, so that the external view remains constant. This is especially true for the "autoRewrite" feature.

I can maybe see a case for `hostRewrite` option always setting the specified value regardless of the redirected target. If you were going to do that, however, I would suggest putting it under `autoRewrite` instead of a fall-through case, since it isn't really a third option. ...And I would add back the proper test cases and add new ones. =)

jcrugzz:
@whatknight I'm with @matthauck, make it an explicit behavior

wingrunr21:
@matthauck The issue I take with with your change is that you are viewing this proxy just like `mod-proxy`. The ability to create custom server logic means that there's no longer a 1-1 mapping of frontend-backend. A developer has the flexibility to build the backend target on the fly (as we have done).

I also take issue with the fact that your change A) silently fails and B) is completely undocumented (same with `autoRewrite`, but that's outside this scope).

I think our use case and the issues we are seeing are getting lost in translation.

@whatknight and I are on a team that's developing a single-page application. During development, the app is served on `localhost`. As we need to interact with backend APIs on different domains, we are using this proxy on the server side to proxy XHR calls out to the various APIs. Custom proxy logic allows us to extract the intended endpoint via a parameter and set that as the proxy target.

As `Host` is `localhost` for these requests, we were manually building the correct `redirectHost` and utilizing the `hostRewrite` functionality to ensure things stayed in line.

We can solve this problem by rewriting the `Host` header to the `target` host. This is reasonable enough, except that you are technically supposed to set `X-Forwarded-Host` to the original `Host` value. Frameworks like Rails, however, will default to using `X-Forwarded-Host` as the base for the `Location` header, which leaves us back at square 1.

matthauck:
Granted, node-http-proxy is not mod-proxy, but it is still a reverse http proxy. I believe the constraint I added is valid for something functioning as a reverse proxy.

I don't believe the constraint I added prevents you from doing what you want to do. The basic reason for the constraint is that a reverse proxy should rewrite `Location` header responses only for the things it is proxying. If the thing it is proxying to sends a redirect to a third-party site, which site is not being reverse-proxied, the proxy should not rewrite the URL in such a way that would cause the client to try to request it via the proxy, which request would fail.

**Scenario 1**: Here's the workflow in my mind for your scenario:

    1. Req(Client --> Proxy): http://localhost/site1/blah
    2. Req(Proxy --> Site1): http://site1.com/blah
    3. Resp(Site1 --> Proxy): HTTP 302, Location: http://site1.com/foo
    4. Resp(Proxy --> Client): HTTP 302, Location: http://localhost/site1/foo

This is how your reverse proxy should work in my opinion. This allows the host constraint to stand. It would need some more nuanced handling of paths, however, to support mapping "/foo" to "/site1/foo". That is the change I think you should make. Or, to make it super-customizable, change it so that it can take a callback function as a config parameter, and do whatever you want in your app for the host redirect.

**Scenario 2**: Here's a workflow that would violate my constraint and which I'm trying to avoid:

    1. Req(Client --> Proxy): http://localhost/site1/blah
    2. Req(Proxy --> Site1): http://site1.com/blah
    3. Resp(Site1 --> Proxy): HTTP 302, Location: http://some-other-site/whatever
    4. Resp(Proxy --> Client): HTTP 302, Location: http://some-other-site/whatever

Note step 4 there, it should not rewrite that redirect since it did not just proxy to that domain. 


Granted, my solution was imperfectly done and probably was poorly documented, but it helped node-http-proxy behave more naturally w/ less configuration as a reverse proxy (rather than hard-coding a global `hostRewrite`, and also handling ports, if memory serves correctly), and in this scenario, the controversial constraint was necessary to add and fixed a bug that was happening for me when scenario 2 above became a reality. 

All that said, I'm not working on this project anymore, and so I don't have much vested interest in keeping the behavior as it is today, and I don't desire much to keep arguing for it. But I am still convinced it is the correct approach. You guys do what you think is best.

PR #850, opened 2015-07-06
Manipulate or cancel proxyRes

cmawhorter:
It'd be nice to have the ability to manipulate or abort a proxyRes.  This is a solution for #737.

Not the prettiest but is 100% backwards compat and has almost [non-existent perf penalty](https://gist.github.com/cmawhorter/2fb412a799f474b4e721).  Test included.

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -164,7 +164,7 @@ web_o = Object.keys(web_o).map(function(pass) {
       }
     });
 
-    proxyReq.end();
+    //proxyReq.end();
   }
 
 ] // <--
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -144,20 +144,27 @@ web_o = Object.keys(web_o).map(function(pass) {
     (options.buffer || req).pipe(proxyReq);
 
     proxyReq.on('response', function(proxyRes) {
+      proxyRes.abort = function() {
+        proxyRes.aborted = true;
+        proxyReq.abort();
+        proxyRes.abort = function(){};
+      };
       if(server) { server.emit('proxyRes', proxyRes, req, res); }
-      for(var i=0; i < web_o.length; i++) {
-        if(web_o[i](req, res, proxyRes, options)) { break; }
+      if (!proxyRes.aborted) {
+        for(var i=0; i < web_o.length; i++) {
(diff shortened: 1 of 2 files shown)


cmawhorter:
Guess I should include an example:

```javascript
proxy.on('proxyRes', function(proxyRes, req, res) {
  // conditionally proxy to a different target 
  if (proxyRes.statusCode === 404) {
    proxyRes.abort(); // leave http-proxy handling
    proxy.web(req, res, { target: 'http://a-different.example.com' });
  }
  // respond directly
  else if (proxyRes.statusCode === 301) {
    proxyRes.abort();
    var someStream = ...;
    someStream.pipe(res);
    res.end();
  }
});
```

prust:
+1  we needed this in order to transparently follow 307 redirects in the proxy for the client

prust:
@indexzero, @jcrugzz: I can rebase this on the latest master if you guys are open to merging it.

euprogramador:
Guys,

I think exactly like that, but I would like to have control over the continuation of the flow as between sending the response from the server to the client I want to run some asynchronous code.

acanimal:
@euprogramador I don't know if it is exactly what you need, but some time ago I was working on ClydeIO project (now stopped due I have no free time to spent on it) which allows to configure a set of middlewares before/after sending request to http-proxy. See image: https://github.com/clydeio/clydeio/wiki/Data-Workflow

thetrevdev:
This is super useful.  Sad to see it wasn't merged

sibnerian:
👍 Among other things, this is extremely useful for catching a proxy error page and sending a prettier version. Would love to see this merged!

For posterity: I was able to hack around it by overriding the ServerResponse methods after sending the response.

```js
    onProxyRes(proxyRes, req, res) {
      if (proxyRes.statusCode >= 400) {
        renderErrorResponse(req, res);
        // Overwrite the methods in https://nodejs.org/api/http.html#http_class_http_serverresponse
        ['addTrailers', 'end', 'setHeader', 'write', 'writeContinue', 'writeHead'].forEach((k) => {
          res[k] = noop;
        });
      }
  }
```
This is a hack, so caveat emptor. It won't work if renderErrorResponse is asynchronous. I'll be getting rid of this when (if?) this PR is merged.

cmawhorter:
anyone still in need of a nodejs proxy should look at anyproxy.  it has a ton of its own problems, but is at least actively maintained by alibaba.  it's perfect for dev, which is all i ever used node proxy for anyway.

Issue #851, opened 2015-07-08
changeOrigin not in documentation


Morriz:
This cost me many hours to figure out....I might be stoopid, but more ppl like me could lost valuable time over this.

Why isn't this option enabled by default? It would serve 99% of the use cases out there.

jpatters:
I also just wasted several hours on this. 
+1 for adding it to the docs

jcrugzz:
I will definitely take a pull-request for it :)

Morriz:
Sure. Who will put that extra line in the docs? Last one in line? I will do it when I have the time ;)

jpatters:
I'd do it but I have no idea where it should go. Any suggestions? The readme is kind of all over the place on doesn't really document anything. It just gives examples. Should it be its own example?

Morriz:
Just mention it in the list of options with a short explanation. Thats enough

bdefore:
:+1: also lost quite a bit of time until i found this in the docs. i'm not sure why it's not true by default!

kdepp:
+1, changeOrigin should be defaultly set to true

indexzero:
@bdefore @kdepp that is not the purpose of this issue. It should have been closed with #852. If you feel strongly about `changeOrigin` please open a new issue if one does not exist. 

I am not opposed to it.

PR #852, opened 2015-07-09
Added missing configuration options

jpatters:
Added missing config options for httpProxy.createProxyServer. Updated to include all options listed in `lib/http-proxy.js`
Addresses nodejitsu/node-http-proxy#851


jcrugzz:
Thanks!

PR #871, opened 2015-08-27
Use raw headers instead parsed.

ArGGu:
Set-Cookie headers are parsed into single header with cookies in array.
This messes up the Set-Cookie headers, because browser receives multiple Set-Cookie headers as single with cookies separted with comma.

### DIFF
--- a/lib/http-proxy/passes/ws-incoming.js
+++ b/lib/http-proxy/passes/ws-incoming.js
@@ -1,5 +1,6 @@
 var http   = require('http'),
     https  = require('https'),
+    util   = require('util'),
     common = require('../common'),
     passes = exports;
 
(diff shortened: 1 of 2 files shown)


indexzero:
Would need a test to verify this is fixing a bug.

ArGGu:
You mean a test application to verify that it fixes the bug or what?

indexzero:
A [mocha-based test](https://github.com/nodejitsu/node-http-proxy/blob/master/test/lib-http-proxy-passes-web-incoming-test.js#L61-L84) which assets that the correct headers are written. Based on your comment the test would assert that a single `set-cookie` header is returned and not multiple, correct?

ArGGu:
Yes browser would receive single set-cookie header with cookies separated with comma.
Because the set-cookie headers are parsed into single header with cookies in array.

There is no problem when server sends a single cookie, but this won't work when server sends multiple cookies.

So if raw headers are used instead parsed, then browser will received all set-cookie headers sent by the server.

This is the first time I have heard of mocha, I will check it out on the weekend.

indexzero:
Alright. Thanks. The other tests in the code should point you in the right direction.

ArGGu:
I added test to lib-http-proxy-test.js.
But I don't understand why the rawHeaders is undefined  when Travis Ci, runs the test.

ArGGu:
http.IncomingMessage in node 0.10 does not have rawHeaders, so it won't run the test with it.

ArGGu:
If it needs to work with node 0.10, then I guess the best solution would be to loop the headers and generate array with checking if the header name is set-cookie.

And if it is then looping the set-cookie array and populate the new array with all set-cookie headers.

jcrugzz:
Yes this needs to work with `0.10.x` still

ArGGu:
I did quick fix and accidently used Object.keys().map instead for in loop.

But yeah it now loops the parsed headers and checks if the header value is array.

If it is, it loops the array and pushes all the header values with the header name to writeHead array.
If not then just pushes the header name with header value.

jcrugzz:
@ArGGu cherry-picked :). Thanks!

PR #883, opened 2015-09-18
resolves #882

### LINKED ISSUE #882: Event to modify outgoing web socket handshake?
I have been working with web sockets in node-http-proxy and I didn't see any path that would allow one to modify the HTTP headers of the proxy request used to conduct the web socket handshake.

The web socket protocol allows the server to fail the connection and make application level decisions based on any other part(s) of the HTTP request (i.e. pre-handshake-response authentication that is not supplied from the incoming request in my case).

It would seem that emitting an event named 'proxyWsReq' with the proxyReq as arguments would be a valid use case.  This could be done right before proxyReq.end() is called in the ws-incoming.js passes.

andrewpmartinez:
Adds a proxyWsReq event that can be used to modify the outgoing
websocket handshake.

### DIFF
--- a/test/lib-http-proxy-passes-ws-incoming-test.js
+++ b/test/lib-http-proxy-passes-ws-incoming-test.js
@@ -1,5 +1,7 @@
 var httpProxy = require('../lib/http-proxy/passes/ws-incoming'),
-    expect = require('expect.js');
+    expect = require('expect.js'),
+    http = require('http'),
+    httpProxyLib = require('../lib/http-proxy');
 
 describe('lib/http-proxy/passes/ws-incoming.js', function () {
   describe('#checkMethodAndHeader', function () {
@@ -9,7 +11,7 @@ describe('lib/http-proxy/passes/ws-incoming.js', function () {
         method: 'DELETE',
         headers: {}
       },
-      stubSocket = { 
+      stubSocket = {
         destroy: function () {
           // Simulate Socket.destroy() method when call
           destroyCalled = true;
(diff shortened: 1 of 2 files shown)


test/lib-http-proxy-passes-ws-incoming-test.js
```diff
@@ -74,6 +76,62 @@ describe('lib/http-proxy/passes/ws-incoming.js', function () {
...
+      });
+
+      function requestHandler(req, res) {
+        proxy.web(req, res);
+      }
+
+      var proxyServer = http.createServer(requestHandler);
+
+      proxyServer.on('upgrade', function(req, res, head){
+          proxy.ws(req, res, head);
```
indexzero: 2-spaces here.

test/lib-http-proxy-passes-ws-incoming-test.js
```diff
@@ -74,6 +76,62 @@ describe('lib/http-proxy/passes/ws-incoming.js', function () {
...
+      });
+
+      var source = http.createServer(function(req, res) {
+          res.end();
```
indexzero: 2-spaces here.

test/lib-http-proxy-passes-ws-incoming-test.js
```diff
@@ -74,6 +76,62 @@ describe('lib/http-proxy/passes/ws-incoming.js', function () {
...
+      });
+
+      source.on('upgrade', function(req, res, head){
+          expect(req.headers['x-special-proxy-header']).to.eql('foobar');
+          source.close();
+          proxyServer.close();
+          done();
```
indexzero: 2-spaces here.

test/lib-http-proxy-passes-ws-incoming-test.js
```diff
@@ -74,6 +76,62 @@ describe('lib/http-proxy/passes/ws-incoming.js', function () {
...
+      proxyServer.listen('8081');
+      source.listen('8080');
+
+      var request = http.request({
+        host: '127.0.0.1',
+        port: '8081',
+        method: 'GET',
+        path: '/',
+        headers: {
+            'upgrade': 'websocket',
```
indexzero: 2-spaces here.

test/lib-http-proxy-passes-ws-incoming-test.js
```diff
@@ -1,5 +1,7 @@
 var httpProxy = require('../lib/http-proxy/passes/ws-incoming'),
-    expect = require('expect.js');
+    expect = require('expect.js'),
+    http = require('http'),
+    httpProxyLib = require('../lib/http-proxy');
```
indexzero:
```
httpProxy = require('../lib/http-proxy');
```

indexzero:
Thank you for your contribution. I am +1 on this once the minor style things are addressed.

andrewpmartinez:
Style issues addressed.

The last change require changing the previous 'var httpProxy' to 'var wsPasses'. Name chosen after seeing the 'var webPasses' for the similar include in the web tests.

andrewpmartinez:
If there is anything else you need done before this is accepted, just let me know.

andrewpmartinez:
When is the next release scheduled?

indexzero:
@donasaur @jcrugzz thoughts on this?

donasaur:
stylistically it looks good. i'm not familiar with this websocket handshake that takes place

indexzero:
@donasaur it's actually not about the Websocket handshake itself. It is mirroring similar `proxyRes` functionality which I am fine with. See https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L147

@jcrugzz thoughts?

donasaur:
ah ok @indexzero , actually it looks like jarrett already merged in a similar PR already:
https://github.com/nodejitsu/node-http-proxy/pull/897/files

can you double check?

indexzero:
@donasaur excellent. Then this is a duplicate of #897. Closing :100:

PR #893, opened 2015-10-18
Removed unspecified trailing slash in proxy url

donasaur:
Previously, if target='http://localhost:9060/appEndpoint' and req.url='', the url the proxied request will hit is: 'http://localhost:9060/appEndpoint**/**'

With the change, if target='http://localhost:9060/appEndpoint' and req.url='', the url the proxied request will hit is: 'http://localhost:9060/appEndpoint'

This is helpful if the app endpoint is to a router that is sensitive to the trailing slash (e.g., a Rails router)


indexzero:
+1

Issue #881, opened 2015-09-17
WS continuation frames & node > 4.0


pepone:
After upgrade to nodejs 4.0.0 I see that the proxy is not forwarding WS trafict that contains continuation frames, tested with nodejs 4.0.0 and 4.1.0 and see the same behavior with both. Everything works fine with node 0.12.7.

Is there any incompatibility issues with node > 4.0?

jedverity:
We're seeing the same issue. Works fine on 0.12.7 but not in iojs or node 4.x.

indexzero:
@pepone @jedverity seems like it. Could you provide a complete console output of what you're seeing? Is there a reproducible test case that you've found? This will help us narrow down the problem.

jcrugzz:
yea this would be super helpful in troubleshooting. I know things related to `http` changed in `iojs` and `node@4` so this could make sense.

jedverity:
Thanks @indexzero @jcrugzz - we're working on putting together a reproducible test case for you right now. Stay tuned.

jcrugzz:
@jedverity awesome :). A failing test PR is my favorite thing

jedverity:
@jcrugzz I'll see what I can do about a failing test. In the meantime, here is a reproducible set-up: https://github.com/jedverity/http-proxy-test

jedverity:
Think I have a (kludgy) test for you @jcrugzz @indexzero . Coming in 5 mins.

jedverity:
@indexzero @jcrugzz @pepone I've confirmed this is fixed in node v5. See http fixes section here: http://nodejs.org/en/blog/release/v5.0.0/

jcrugzz:
@jedverity hmm well we should see if whatever fix can be backported if it is not a breaking change. Have you tried it in `4.2.1`? That is the LTS release so we should get this fixed in that release

jedverity:
@jcrugzz I did try it in 4.2.1 and no dice. Same results as other 4.x versions. I assumed this was the fix but haven't evaluated closely: https://github.com/nodejs/node/pull/3342

jcrugzz:
@jedverity i think you might be right, looks like its on target for `4.2.2` if you are correct. See if you can test the `4.2.2` release candidate for confirmation. Either way your tests are still good :). Will add them anyway

jedverity:
@jcrugzz Confirmed 4.2.2 works

PR #901, opened 2015-10-28
Add tests for forwarding of continuation frames

jedverity:
This adds two tests that send payloads below and at the threshold for continuation frames. Using node 0.12.7 both tests pass. Using node 4.1.2 the test below the threshold passes but the other fails. Prompted by #881 .


jedverity:
Note: This issue appears to have been fixed in node v5. See http fixes section here: http://nodejs.org/en/blog/release/v5.0.0/

PR #910, opened 2015-11-21
Updated markdown docs to mention proxy rules module

donasaur:


indexzero:
@donasaur could you add some code samples as well? I think that will help a lot of users better understand what they are getting from `proxy-rules`.

_Related:_ we should do a pass at updating the `README.md` sometime in the next few cycles. For starters: a Table of Contents that highlights this feature more prevalently would help a lot of users as well.

donasaur:
Sure. For the former, should I do something like list a more readable version of these [tests](https://github.com/donasaur/http-proxy-rules/blob/master/test/index.tests.js#L38-L64) on the `proxy-rules` README?

indexzero:
@donasaur yeah that sounds good.

donasaur:
@indexzero Kk, I updated the README with a link to the relevant portion of the test, and made that portion readable with comments and a little name refactoring. See [here](https://github.com/donasaur/http-proxy-rules/commit/16d60201a929d6a3a8ffa3a5ef9f856490b7a272)

jcrugzz:
Looks good :+1:. I agree this readme could use another walk through and have a proper table of contents. Good next steps

PR #922, opened 2015-12-07
SSE example and test

aaronmaturen:


jcrugzz:
Awesome thank you @aaronmaturen

Issue #915, opened 2015-11-28
node-http-proxy SSL connections is not working. 


BadBoy20:
I have set up a proxy node-http-proxy server on port 8009. and my actual http server on port 3000. But it seems like they are not working. My code looks like this:

var fs = require('fs');

var httpProxy = require('http-proxy');

httpProxy.createServer({
  target: 'http://127.0.0.1:3000/',
  ssl: {
    key: fs.readFileSync('./sslcerts/key.pem', 'utf8'),
    cert: fs.readFileSync('./sslcerts/cert.pem', 'utf8'),
    passphrase: 'password1'
  },
  requestCert: true,
  rejectUnauthorized: true
}).listen(8009, function(){
  console.log("SSL server started listening on port 8009");
});

// thats the node http proxy

var http = require('http');

var port = 3000;
app.set('port', port);

var server = http.createServer(app);

server.listen(port, function(){
  console.log("Express non-SSL server listening on port " + port);
});

//thats the http web server itself.
-----------------------------------------

Now I have used wireshark to sniff the traffic and there was a post field that was sending data to the https://localhost:8009/ server and it was viewable in plain text on the localhost loopback interface.
Not only that, the wireshark protocol detected was http (not TLS/SSL like its supposed to show). there was no indication of TLS handshake or anything like that. I have tested with other legitimate webpages with wireshark and what I found was those webpages have a TLS key handshake like the client hello/server hello stuff but that was not seen when I used wireshark on a request to localhost:8009 https server. So I would like to know what I could do to fix the issue and go through the steps of the TLS server hello/client hello handshake and make sure the data is not being sent on plaintext but rather, encrypted.

pyper:
@BadBoy20 I think it is the port.  I had all kinds of issues trying to get it to connect ports that weren’t 80 or 443.  Once I switched to those it worked fine.  You can have your proxy listening on 8009, just trying changing your test server upstream from 3000 to 443.

If it fixes it, it looks the issue is under:

https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/common.js

Selection of port seems to be in multiple places and inconsistent.

BadBoy20:
[...]

PR #924, opened 2015-12-11
Fixes issues with incorrect port selection

pyper:
Hey everyone,

This is in regards to an issue I brought up in #915 yesterday.  I was having an issue with incorrect port selection and assumed @BadBoy20 might be having the issue too.  Anyhow, the problem for me was explicit declaration of the port was not taking precedence over the port implied by the protocol.  The unit test I created was failing before I made the fix.  Also, I did not check the getPort() function in common.js as it’s only purpose is setting “x-forwarded” headers in web-incoming.js and ws-incoming.js so it seemed out of scope for this issue.

Cheers.

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -31,7 +31,8 @@ common.isSSL = isSSL;
  */
 
 common.setupOutgoing = function(outgoing, options, req, forward) {
-  outgoing.port = options[forward || 'target'].port ||
+  outgoing.port = options.port || 
+                  options[forward || 'target'].port ||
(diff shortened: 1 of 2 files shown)


pyper:
Anyone going to merge this?

jcrugzz:
@pyper this adds another option which does handle a case where we want to pass it in explicitly which is great, but I dont think this fully solves the problem discussed. Could you see if you can fix the bug where the port in the `target` URL itself is properly used? Thanks for your help!

indexzero:
Closing due to age and in context of prior discussion indicating it was not a complete fix. Thanks . for your contribution!

Issue #839, opened 2015-06-10
Ignore path and the trailing slash


mereghost:
As I need to rewrite the request URL as it does not map perfectly to the applications under the proxy, I've set ignorePath to true.

Then I proceeded to define the target as a full doctored url (a valid URL for my applications), but as in common.js:82 it does not truly ignore the path. It sets it to '/' and them proceeds to append it to target / forward.

Is there any way around that? Is there a better way to do what I need?

mereghost:
Hmmm... the only current way (that I found) to bypass this behavior is rewriting the proxied request path using the callback 'proxyReq'

```javascript
proxy.on('proxyReq', function (proxyReq, req, res, options) {
    proxyReq.path = path;
});
```

bdefore:
thanks @mereghost i was looking for just this. to prepend, you can do i.e. `proxyReq.path = "/api/" + proxyReq.path'. certainly would be nicer if `ignorePath` was better documented.

cdl:
Thanks @mereghost - just ran into this issue myself, but it didn't surface until dealing with proxying requests with query parameters. Glad to see I'm not going insane. :+1:

RickWong:
:+1:

darrenmce:
:+1: `/` appended to end of querystring when using `ignorePath`

75lb:
i need to rewrite the target path too for the exact same reason (the target URL structure does not match the source URL structure). 

[this appended trailing slash](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/common.js#L91) breaks the ability to rewrite the target path as a specific resource.. could we remove it and update the docs for `ignorePath` stating that, if required, we must manually append our own `target` trailing slash?

niclic:
Apparently resolved by this: https://github.com/nodejitsu/node-http-proxy/pull/934

jcrugzz:
@niclic should be, thanks :)

PR #934, opened 2016-01-13
Fix for #839 (Ignore path and the trailing slash)

Torthu:
When using option ignorePath common.js would set target to '/'.

This would break requests in cases where this slash was unwanted, for instance when fetching files, or in some cases when using GET parameters.


jcrugzz:
Thanks!

Issue #946, opened 2016-01-28
Url corruption when ":/" present


coderaiser:
Why are you replace lines: `:/` to `://` in [urlJoin of common.js](https://github.com/nodejitsu/node-http-proxy/blob/302d981dd2cf06dbf751b1f64e3dfea08d0f9476/lib/http-proxy/common.js#L185)?

It corrupt work of [join-io](https://github.com/coderaiser/join-io) which uses such urls:
```html
<script src="/join:/lib/client.js:/lib/util.js:/join/join.js"></script>
```

jcrugzz:
@coderaiser its meant to handle cases where `http://whatever.com` turns into `http:/whatever.com`. This is a naive urlJoin since it fails on this case, it should be more specific to the post `http`. I would love a PR to fix this if you can turn this into a test case :)

coderaiser:
The thing is no `http://` string is present on args. Here is example
```js
args = [ '/',  '/iocmd/join:/modules/format-io/lib/format.js:lib/client/directory.js:/join/join.js' ] 
retSegs = [ '/iocmd/join://modules/format-io/lib/format.js:lib/client/directory.js://join/join.js' ]
```

jcrugzz:
@coderaiser Yes I understand. I was mentioning thats why it exists, Im not saying its correct. The code here needs to be smarter to handle the case you are mentioning.

PR #947, opened 2016-01-28
fix(common) urlJoin replace: ":/" -> "http?s:/"

coderaiser:
Shrunk replace options.
Fix https://github.com/nodejitsu/node-http-proxy/issues/946.


jcrugzz:
@coderaiser could you add a test for the case you pointed out in #946

coderaiser:
@jcrugzz done

Issue #949, opened 2016-02-02
README example implies request instead of response parameter


nitind:
The repository README implies that the `close` callback will be given a request object, when it's really the proxy response object according to the relevant source.

//
// Listen for the `close` event on `proxy`.
//
proxy.on('close', function (req, socket, head) { // <-- `req` should be `res`
  // view disconnected websocket connections
  console.log('Client disconnected');
});

jcrugzz:
@niallo great find! Would love a quick fix PR :).

caioquirino:
Quick Fix PR created :)

PR #950, opened 2016-02-02
#949 Proxy example using req instead res on README

caioquirino:
README example implies request instead of response parameter


jcrugzz:
Thanks @caioquirino

PR #957, opened 2016-02-10
[fix] pass more useful information to the error that we dont already …

jcrugzz:
So the caveat here is that this is a breaking change, but i think it provides more useful context to the `error` as the `error` should be the actual `proxyReq` and not the `req` itself. In most cases (or at least using the `callback` api, you already have access to `req` and `res` so providing these ends up being useless. We were also previously capturing errors from client requests which I believe was creating false positives for people.

Thoughts @donasaur @indexzero


donasaur:
Yea, that's more helpful. I think if you also update the README for event listening, then it'll be fine

PR #961, opened 2016-02-15
Remove "transfer-encoding" header if "content-length" is set to zero

### LINKED ISSUE #960: Do not use "Transfer-Encoding: chunked" header for proxied DELETE requests (again)
This issue is related to PR #373 where this exact bug was already fixed by a89e2f2, but then later apparently introduced again by 8663ac1.

The actual problem is that `DELETE` requests with both `Content-Length: 0` and `Transfer-Encoding: chunked` are illegal and are being rejected by some servers (in this case nodejs since v0.10.42). If the incoming request only has the `Transfer-Encoding` header the proxy will somehow automatically set the `Content-Length` header to `0` but won't remove the encoding header which leads to this issue.

Turbo87:
This PR resolves #960


jcrugzz:
LGTM Thanks! :+1:

Issue #813, opened 2015-04-27
How can I filter socket hang up errors?


JSteunou:
I'm using node-http-proxy for a while now, and I think I'm using it quite right catching errors like this

```javascript
proxy.on('error', function(err, req, res) {
    // log
    // ...
    // handle
    if (!res.headersSent) {
        res.writeHead(500, { 'content-type': 'application/json' });
    }
    res.end(JSON.stringify({ error: 'proxy_error', reason: err.message }));
});
```

But I can have two cases that make a "socket hang up" error happen.

1. Crash behind
2. Client cancelling the request

Watching at my log I can make no differences between those two cases and it's quite frustrating. Before knowing about request abort (thank #527) I really though I have an error going on.

So, what's the best practice to filter the request.abort case out?

theoephraim:
+1

FoghostCn:
+1

vvo:
You should be able to use the proxy error event or the error callback and check for res.socket.destroyed === true and error.code  === ECONNRESET which means the request was canceled

JSteunou:
Merci @vvo ;) Will try that asap

JSteunou:
It works very fine, thank you again @vvo 

@jcrugzz should I let this issue open? Maybe http-proxy should filter those error inside the library, or maybe it's better to let it unopinionated...

vvo:
> or maybe it's better to let it unopinionated...

Yes it depends on the use case, sometime you will want the error to pop out if the client canceled the request.

JSteunou:
What I thought. Closing it then.

JSteunou:
Quick question again: I want to use [request abort](https://nodejs.org/docs/v0.12.8/api/http.html#http_request_abort) to prevent the underlying server to handle that client cancelled request. But, very surprisingly, `req.abort` is `undefined` O_O

Any thoughts?

vvo:
nope sorry,

JSteunou:
I think the `req` I got is the received request, not the request currently sent by http-proxy

JSteunou:
Ok for the record, in caise anyone was interested at doing the same thing

### How to catch request client cancelled and abort ongoing proxied request 

```
proxy.on('proxyReq', function(proxyReq, req) {
    // keep a ref
    req._proxyReq = proxyReq;
});

proxy.on('error', function(err, req, res) {
  if (req.socket.destroyed && err.code === 'ECONNRESET') {
    req._proxyReq.abort();
  }
[...]

PR #966, opened 2016-02-24
Emit disconnected event instead of error when ECONNRESET

Deividy:
Hello folks, 
This is my first contribution, if there is anything wrong with my PR, please let me know.

We've been using node-http-proxy in our prod server since 3 years now and its working as a charm (thank you!), but everyday we got our log files full of `socket hang up` messages, so I come here dig code and found an issue for that: #[813](https://github.com/nodejitsu/node-http-proxy/issues/813). 

To me, this looks like a bug also, we should not send as an error a disconnection between the client, so I made this PR using the info in the issue.

I'm not sure if emit a `disconnected` event is the best approach, or if you guys want something more elaborate for that. Please, let me know your thoughts.

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -132,7 +132,15 @@ web_o = Object.keys(web_o).map(function(pass) {
     req.on('error', proxyError);
 
     // Error Handler
-    proxyReq.on('error', proxyError);
+    proxyReq.on('error', function (err) {
+        if (req.socket.destroyed && err.code === 'ECONNRESET') {
+          server.emit('disconnected', err, req, res, options.target);
+          proxyReq.abort();
+        } else {
+          proxyError(err);
+        }
+    });
+
 
     function proxyError (err){
       if (clb) {
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -132,17 +132,14 @@ web_o = Object.keys(web_o).map(function(pass) {
     req.on('error', proxyError);
 
     // Error Handler
-    proxyReq.on('error', function (err) {
-        if (req.socket.destroyed && err.code === 'ECONNRESET') {
-          server.emit('disconnected', err, req, res, options.target);
-          proxyReq.abort();
-        } else {
-          proxyError(err);
-        }
-    });
-
+    proxyReq.on('error', proxyError);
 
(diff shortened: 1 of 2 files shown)


jcrugzz:
@Deividy thanks for the contribution! I saw you did some investigation on removing the `req.on('error')` listener. The proxyReq error event was still triggered when the client disconnected when removing it still? I can see this can still be an issue when the client request itself timed out but was curious :). Will comment

lib/http-proxy/passes/web-incoming.js
```diff
@@ -132,7 +132,15 @@ web_o = Object.keys(web_o).map(function(pass) {
     req.on('error', proxyError);
 
     // Error Handler
-    proxyReq.on('error', proxyError);
+    proxyReq.on('error', function (err) {
+        if (req.socket.destroyed && err.code === 'ECONNRESET') {
```
jcrugzz: we should contain this logic in the `proxyError` function and change the error listener for `req` if we are going to keep it

jcrugzz: This should also be `2` spaces not `4` spaces for indentation.

lib/http-proxy/passes/web-incoming.js
```diff
@@ -132,7 +132,15 @@ web_o = Object.keys(web_o).map(function(pass) {
...
+          server.emit('disconnected', err, req, res, options.target);
```
jcrugzz: maybe make the event `econnreset`? its more for logging purposes then anything

Deividy:
Yep, the `proxyReq` error was still triggered even removing the listen `req.on('error')`, my test case was refresh the browser window like crazy, this can trigger the `socket hang up` false alarms.

Thanks for the comments, I made the tweaks. 
About the `req` listener, what we can do about that? Maybe emit some other event or just remove it? I saw a PR removing it, don't know if I touch that or not.

JSteunou:
Maybe missing a line of comment to explain why this function has an early return instead of calling callback or triggering error event. Otherwise looks very good, thank you for this PR!

JSteunou:
@jcrugzz good to go?

jcrugzz:
@Deividy thank you for your work. Will merge this. id like to rebase the branch that removes teh `req.on('error')` listener and see how it behaves.

AnujTewari:
@Deividy and @jcrugzz - can the same be used for `ECONNREFUSED`?

PR #967, opened 2016-02-26
Set the x-forwarded-host flag when xfwd is enabled

jbacklund:
Reasoning: Rack's request class [makes use of](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L243) this HTTP header. Certain edge-case scenarios (proxying from ember-cli to a Rails backend) can be problematic without this header being present.

/cc @perlun, @jesjos

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -78,6 +78,8 @@ web_o = Object.keys(web_o).map(function(pass) {
         (req.headers['x-forwarded-' + header] ? ',' : '') +
         values[header];
     });
+    
+    req.headers['x-forwarded-host'] = req.headers['host'];
   },
 
   /**


jcrugzz:
Seems reasonable.

chimurai:
@jbacklund @jcrugzz 

With this change, the `xfwd` behavior for `web` is slightly different from `ws`.

https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L65-L83
https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/ws-incoming.js#L54-L69

Think this change should be applied to [WebSockets](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/ws-incoming.js#L54-L69
) as well, instead of just the [Web](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L65-L83) part.

perlun:
Agree, we should fix that as well. Care to submit a PR?

PR #974, opened 2016-03-08
Fix formatting of the `headers` option

bmac:
### Before
screen shot 2016-03-08 at 6 06 30 pm

### After
screen shot 2016-03-08 at 6 07 31 pm


jcrugzz:
Thanks!

PR #975, opened 2016-03-11
fix browserify compatibility

idjem:


jfhbrook:
Just a drive-by here cause twitter and my own curiosity:

How does this break/fix browserify?

jcrugzz:
Yea this is interesting that it does fix something but harmless regardless. Will push out a release when I'm home.

PR #1001, opened 2016-05-04
Close the connection when target sends "Connection: close" header

### LINKED ISSUE #1000: "Connection: close" header from target server is not honored
When the target server sends a `Connection: close` header, normal behavior as per the HTTP protocol is that both the client and server should close the connection when data transmission has completed.

Currently node-http-proxy doesn't honor this. Instead the connection is left open, which results in `ECONNRESET` errors being thrown on subsequent requests.

There are several related open issues, but I couldn't find any that directly address the same problem. For example: #446, #813, #859, #757, #966.

nicolayr:
This prevents ECONNRESET errors from being thrown for connections
that are supposed to be closed (fixes #1000).

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -137,6 +137,12 @@ web_o = Object.keys(web_o).map(function(pass) {
     proxyReq.on('error', proxyError);
 
     function proxyError (err){
+      if (err.code === 'ECONNRESET' && proxyReq._headers && proxyReq._headers.connection === 'close') {
+        // The target server has closed the connection, but since it first sent
+        // a 'Connection: close' header this is not an error.
+        return;
+      }
+
       if (clb) {
         clb(err, req, res, options.target);
       } else {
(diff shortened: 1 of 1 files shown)


nicolayr:
The initial change caused some connections to be closed before all data was transmitted to the client.

I changed it so that instead of closing connections, connection errors for connections that are _supposed_ to be closed don't trigger the error callback.

lib/http-proxy/passes/web-incoming.js
```diff
@@ -137,6 +137,12 @@ web_o = Object.keys(web_o).map(function(pass) {
     proxyReq.on('error', proxyError);
 
     function proxyError (err){
+      if (err.code === 'ECONNRESET' && proxyReq._headers && proxyReq._headers.connection === 'close') {
```
jcrugzz: im about to merge the case with the `econnreset` throwing errors that arent really errors. you should rebase all that and try and get the tests to pass and ill consider the use case here.

nicolayr:
Seems to have been fixed by #966

PR #1009, opened 2016-05-18
Add option to rewrite domain in set-cookie headers

Volune:
Sometime the `set-cookie` header specify a domain for the cookie. When proxying the a target using `changeOrigin` option, the response may contain a cookie not matching the domain of the client.

Example of such a cookie: `foo=bar; domain=my.domain; expires=Tue, 12-Feb-2019 10:50:41 GMT; path=/`

This pull request adds an option to rewrite the `set-cookie` header of the response.

Possible option values:
- disabled (default): `cookieDomainRewrite: false`
- rewrite all domains: `cookieDomainRewrite: "my.client.domain"`
- remove all domains: `cookieDomainRewrite: ""`
- more advanced configuration (this example removes all except one):

  ```
  cookieDomainRewrite: {
    "*": "",
    "some.domain": "some.domain"
  }
  ```

Let me know if I can improve this option or for any question.

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -4,7 +4,8 @@ var common   = exports,
     required = require('requires-port');
 
 var upgradeHeader = /(^|,)\s*upgrade\s*($|,)/i,
-    isSSL = /^https|wss/;
+    isSSL = /^https|wss/,
+    cookieDomainRegex = /(;\s*domain=)([^;]+)/i;
 
 /**
  * Simple Regex for testing if protocol is https
@@ -199,6 +200,41 @@ common.urlJoin = function() {
   retSegs.push.apply(retSegs, lastSegs);
 
   return retSegs.join('?')
+};
+
+/**
+ * Rewrites or removes the domain of a cookie header
+ *
+ * @param {String|Array} Header
+ * @param {Object} Config, mapping of domain to rewritten domain.
+ *                 '*' key to match any domain, null value to remove the domain.
+ *
+ * @api private
+ */
+common.rewriteCookieDomain = function rewriteCookieDomain(header, config) {
+  if (Array.isArray(header)) {
+    return header.map(function (headerElement) {
+      return rewriteCookieDomain(headerElement, config);
+    });
+  }
+  return header.replace(cookieDomainRegex, function(match, prefix, previousDomain) {
+    var newDomain;
+    if (previousDomain in config) {
+      newDomain = config[previousDomain];
+    } else if ('*' in config) {
+      newDomain = config['*'];
+    } else {
+      //no match, return previous domain
+      return match;
+    }
+    if (newDomain) {
+      //replace domain
+      return prefix + newDomain;
+    } else {
+      //remove domain
+      return '';
+    }
+  });
 };
 
 /**
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -1,4 +1,5 @@
 var url    = require('url'),
+    common = require('../common'),
     passes = exports;
 
 var redirectRegex = /^201|30(1|2|7|8)$/;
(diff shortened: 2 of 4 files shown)


drouillard:
Looks promising

parse:
Would love to see this merged.

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -80,12 +80,44 @@ var redirectRegex = /^30(1|2|7|8)$/;
...
+      var header = proxyRes.headers[key];
+      if(header != undefined){
+        if(rewriteCookieDomainConfig && key.toLowerCase() === 'set-cookie') {
+          header = rewriteCookieDomain(header);
+        }
+        res.setHeader(String(key).trim(), header);
       }
     });
+
+    function rewriteCookieDomain(header) {
```
jcrugzz: this function should move to the `common.js` file as it is not a `pass` that accepts the `(req, res, ...)` signature.

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -80,12 +80,44 @@ var redirectRegex = /^30(1|2|7|8)$/;
    *
    * @api private
    */
-  function writeHeaders(req, res, proxyRes) {
+  function writeHeaders(req, res, proxyRes, options) {
+    var rewriteCookieDomainConfig = options ? options.cookieDomainRewrite : undefined;
+    if(typeof rewriteCookieDomainConfig === "string") { //also test for ""
```
jcrugzz: single quotes, space after `if`.

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -80,12 +80,44 @@ var redirectRegex = /^30(1|2|7|8)$/;
    *
    * @api private
    */
-  function writeHeaders(req, res, proxyRes) {
+  function writeHeaders(req, res, proxyRes, options) {
+    var rewriteCookieDomainConfig = options ? options.cookieDomainRewrite : undefined;
```
jcrugzz: no need for ternary operator here. `var cookieDomainRewrite = options.cookieDomainRewrite`

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -80,12 +80,44 @@ var redirectRegex = /^30(1|2|7|8)$/;
...
+    if(typeof rewriteCookieDomainConfig === "string") { //also test for ""
+      rewriteCookieDomainConfig = {"*": rewriteCookieDomainConfig};
```
jcrugzz: single quotes and add spaces around `{ '*': cookieDomainRewrite }`

lib/http-proxy/passes/web-outgoing.js
jcrugzz: This should actually be its own module but for now you can move it to `common.js` and change the function signature to include the defined options you need.

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -80,12 +80,44 @@ var redirectRegex = /^30(1|2|7|8)$/;
...
+
+    function rewriteCookieDomain(header) {
+      if(Array.isArray(header)) {
+        return header.map(rewriteCookieDomain);
+      }
+      return header.replace(/(;\s*domain=)([^;]+)/, function(match, prefix, previousDomain) {
+        var newDomain;
+        if(previousDomain in rewriteCookieDomainConfig) {
+          newDomain = rewriteCookieDomainConfig[previousDomain];
+        } else if ("*" in rewriteCookieDomainConfig) {
```
jcrugzz: single quotes everywhere here

Volune:
PR should be clean now, I also rebased on top of current master, and improved function comments. Let me know if I missed anything.

lib/http-proxy/common.js
```diff
@@ -198,6 +198,41 @@ common.urlJoin = function() {
...
+ *
+ * @api private
+ */
+common.rewriteCookieDomain = function rewriteCookieDomain(header, config) {
+  if (Array.isArray(header)) {
+    return header.map(function (headerElement) {
+      return rewriteCookieDomain(headerElement, config);
+    });
+  }
+  return header.replace(/(;\s*domain=)([^;]+)/, function(match, prefix, previousDomain) {
```
jcrugzz: Lets cache this regex so it isnt created every time the function is called. Put `var domainRegex = ...` up top here and use it in this function.

jcrugzz:
@Volune Overall this looks good. I would like to see some more documentation for this option in the README, similar to what you have in this PR in your initial proposal. The only thing that kind of bothers me is having `''` remove all domains. Would it be more intuitive to use an explicit `null` as removal? I'm really not sure what the right answer is but is there something that semantically makes more sense for that implication?

Volune:
@jcrugzz The first reason of using `''` is that it would transform `Domain=some.domain` to `Domain=`, which has the meaning of no domain, but (as far as I understand RFC 6265) is not a valid value. So I used it to remove the domain.

Also, it adds some meaning to the type of the option:
- boolean `false`: disable the option
- object: advanced configuration
- string: shorthand for `{ '*': value }` configuration

I'm not against using explicit `null` and/or explicit `undefined` to remove the domain, and document it.
Let me know what's your opinion, decision.

jcrugzz:
@Volune Thats reasonable, i think its ok. Lets just document all the options for this option in the readme, take care of that one nit i have so that we are a bit more performant so we arent creating the regex everytime and we cache it in a variable at the top of `common.js` (with the other one thats there. Then we will get this merged :). Thanks for bearing with me

jcrugzz:
Also we just need a minor rebase. since i merged some other PRs

Volune:
Should be all good.

isaachinman:
Hey everyone, thanks for this - a very helpful feature. 

Wondering if there is any built-in way to also rewrite secure/httpOnly option? In local development, I am using `localhost:3000`, which obviously isn't secure, so even with the domain rewrite, cookies are failing to be set.

joeskeen:
@isaachinman I am faced with this same issue as well.  Perhaps it would be good to open up a separate issue for this feature.

eonlepapillon:
@isaachinman Just information, node-proxy-middleware has this implemented. It can't be hard to implement this in node-http-proxy

https://github.com/gonzalocasas/node-proxy-middleware/blob/master/index.js#L113

PR #1010, opened 2016-05-24
Fix tests for node 0.10 + socket.io-client@1.4.6 (engine.io-client@1.6.9)

Volune:
One test "`should emit open and close events when socket.io client connects and disconnects`" is broken since socket.io-client upgraded to 1.4.6 (upgrading its dependency engine.io-client to 1.6.9), because of a new default value in commit: socketio/engine.io-client@7fa16d531b40fb316e418e8f3d9e50bb07b38531

This pull request force the value of `rejectUnauthorized` to null (like before upgrade) to fix the test.


jcrugzz:
Thanks so much! 👍

PR #1018, opened 2016-06-02
Conserve headers' case

diegonc:
Since version [1.0.0](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_IOJS.md#http), Node has added support for getting the raw headers from an incomming message, like the ones `http.request` passes to its callback.

This opens the possibility of sending the headers to the client using the same capitalization as the target server used when responding to our request.

What the PR proposes to accomplish this goal is to add a new option, `keepHeaderCase`, that instructs the `writeHeaders` outgoing pass to build a map of lower-cased headers to its original representation located in the `rawHeaders` property of `proxyRes`.

To keep backwards compatibility with earlier versions of Node, the outgoing pass above uses the keys of the `proxyRes.headers` map as the first parameter of `res.setHeader`.

There are some corner cases regarding headers set by preceding passes; it wasn't clear to me what to do with those. But otherwise, the simple test added by the first commit in this PR works ok.

What do you think about this feature?

Thanks,
Diego

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -80,10 +80,26 @@ var redirectRegex = /^30(1|2|7|8)$/;
    *
    * @api private
    */
-  function writeHeaders(req, res, proxyRes) {
+  function writeHeaders(req, res, proxyRes, options) {
+    function collectHeaderNames (rawHeaders) {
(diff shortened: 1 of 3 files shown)


diegonc:
Hello, I'm closing this as result of doing cleanup of my PRs. I believe this feature was already implemented by @pachirel so this patch is not required anymore.

Thanks

Issue #914, opened 2015-11-24
`rewrite` options should apply to all responses


tillsc:
According to the documentation the `hostRewrite` and `autoRewrite` options only apply to 3xx responses. But the `Location`-Header is allowed in more than only the 3xx responses (e.g. in `201 Created`).

I don't understand why it is necessary to limit this functionality to specific response codes. Why couldn't just every `Location` header be analyzed and replaced if there is one present in the response?

gabrielboucher:
I agree, this is currently biting me. We are implementing an API that tries hard to follow this specification: http://jsonapi.org/format/#crud-creating-responses especially where it says "The response SHOULD include a Location header identifying the location of the newly created resource."

However this request returns a 201 status code and the location is not being rewritten.

jrnail23:
related: #845

PR #1024, opened 2016-06-13
Location header rewriting for responses with status 201

gabrielboucher:
Implement rewriting of the location header for responses with status code 201, according to RFC2616 section 10.2.2, in response to issue #914


jcrugzz:
@gabrielboucher seems reasonable to me if its in the spec. Thanks for the contribution!

gabrielboucher:
The test is already there

jcrugzz:
ha, looking too quickly. thanks for keeping me on my toes :)

Issue #843, opened 2015-06-26
node-http-proxy XHR post call hanging


elankeeran:
Create a simple express app and try to do proxy. 

Proxy working fine when request is GET but it getting hang on post XHR call

```javascript
var express = require('express');
var path = require('path');
var logger = require('morgan');
var cookieParser = require('cookie-parser');
var bodyParser = require('body-parser');
var httpProxy = require('http-proxy');
var app = express();
app.use(bodyParser.json());
app.use(bodyParser.urlencoded({ extended: false }));
app.use(cookieParser());
var apiProxy = httpProxy.createProxyServer();
app.use(function(req, res,next){
    apiProxy.web(req, res, { target: 'http://localhost:9000' } );

});
// production error handler
// no stacktraces leaked to user
app.use(function(err, req, res, next) {
  res.status(err.status || 500);
  res.render('error', {
    message: err.message,
    error: {}
  });
});

module.exports = app;
```

Googled and got this solution https://github.com/nodejitsu/node-http-proxy/issues/180

```javascript
var restreamer = function (){
        return function (req, res, next) { //restreame
            req.removeAllListeners('data')
            req.removeAllListeners('end')
            next()
            process.nextTick(function () {
                if(req.body) {
                    req.emit('data', JSON.stringify(req.body))
                }
                req.emit('end')
            })
        }
    }

    app.use(restreamer());
```

but now application not hanging. I am getting response 404 not found. and getting below error

```error
/Volumes/Data/nodejs/express-apps/myapp/node_modules/http-proxy/lib/http-proxy/index.js:119
    throw err;
          ^
Error: socket hang up
    at createHangUpError (_http_client.js:215:15)
    at Socket.socketOnEnd (_http_client.js:300:23)
    at Socket.emit (events.js:129:20)
    at _stream_readable.js:908:16
    at process._tickCallback (node.js:355:11)
```

WHK102:
Same problem with reverse proxy and jQuery ajax post call:

/home/whk/Escritorio/node/node_modules/http-proxy/lib/http-proxy/index.js:119
    throw err;
          ^
Error: socket hang up
    at createHangUpError (http.js:1473:15)
    at Socket.socketCloseListener (http.js:1523:23)
    at Socket.emit (events.js:95:17)
    at TCP.close (net.js:466:12)

prashantbaid:
[...]

Issue #791, opened 2015-03-11
body parser before proxy request


nguyenxuantuong:
I am wondering is there anyway to use body parse before proxy request. Here is my scenario: 
app.use(bodyParser.json())
app.use(bodyParser.urlencoded({ extended: false }));

proxy.on('proxyReq', function(proxyReq, req, res, options) {
    //Need to write custom logic to check for the parameter inside the body. -- so that it can add custom header or modify proxyReq body, etc...
    //for example: var userId = req.param("userId") -> this one only available with bodyParser declared   //before this one (see abover) 
});
app.all(route, function(req, res){ 
       //proxy request to target url
        return proxy.web(req, res, { target: config.localUrl + "/api"}, function(e) {
            return res.json(501, e.stack || e.error || e);
        });
    });

The web application at target also uses its own bodyParser. But it seems that because the body parser being used twice, no response arrived after proxy the request to target.  I am wondering if there is any way to work around that issue? Thank.

lchenay:
Please see : https://github.com/nodejitsu/node-http-proxy/blob/master/examples/middleware/bodyDecoder-middleware.js

PR #1027, opened 2016-06-23
[Fix] bodyparser example - restream body before proxying

### LINKED ISSUE #955: POST proxying with bodyParser middleware
The following is my code.

**route.js**

```
var express = require('express');
var httpProxy = require('http-proxy');
var bodyParser = require('body-parser');
var proxy = httpProxy.createProxyServer({secure:false});
var app = express();
var jsonParser = bodyParser.json();

proxy.on('proxyReq', function(proxyReq, req, res, options) {
    logger.debug("proxying for",req.url);
    //set headers
    logger.debug('proxy request forwarded succesfully');
});

proxy.on('error', function (err, req, res) {
  res.writeHead(500, {
    'Content-Type': 'text/plain'
  });
  res.end('Something went wrong. And we are reporting a custom error message.');
});

proxy.on('proxyRes', function (proxyRes, req, res) {
  console.log('RAW Response from the target', JSON.stringify(proxyRes.headers, true, 2));
});

module.exports = function(app){
  app.post('/recording',jsonParser,function(req,res){
    // update request body
    proxy.web(req, res, { target: <<host>>:<<port>>});
  });
}
```

**app.js**

```
var express = require('express');
var app = express();
require('./routes')(app);

app.listen(8080);
console.log("Demo server running");
```

But still the POST request hangs and ultimately fails.

I also use bodyparser middleware and it has a known issue as mentioned in [Github](https://github.com/nodejitsu/node-http-proxy/issues/180) issue. So I tried adding this line as the last line in app.js

I also tried restreamer code but I get the following error if I use the restreamer

`error: uncaughtException: Can't set headers after they are sent.
`
I get this error because I emit end in the restreamer and then try to set the headers in proxy.on function. Any suggestions appreciated.

matibek:
Support for bodyparser.json and bodyparser.urlencoded.

Fixes #955 #843 #791

### DIFF
--- a/examples/middleware/bodyDecoder-middleware.js
+++ b/examples/middleware/bodyDecoder-middleware.js
@@ -34,28 +34,25 @@ var http = require('http'),
     proxy = httpProxy.createProxyServer({});
 
 
-//restreame
-var restreamer = function (){
-  return function (req, res, next) { //restreame
-    req.removeAllListeners('data')
-    req.removeAllListeners('end')
-    next()
-    process.nextTick(function () {
-      if(req.body) {
-        req.emit('data', JSON.stringify(req.body))
-      }
-      req.emit('end')
-    })
+//restream parsed body before proxying
+proxy.on('proxyReq', function(proxyReq, req, res, options) {
+  if(req.body) {
+    let bodyData = JSON.stringify(req.body);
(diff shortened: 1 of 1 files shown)


jcrugzz:
Thanks! good looks.

PR #1031, opened 2016-07-06
close the proxy connection to avoid ECONNRESET on windows

FredericHeem:
On Windows, when simultaneous connections are proxied, some of the connections are closed with a `ECONNRESET`. This PR ensures that the connection to the remote is closed.


FredericHeem:
Unfortunately, this PR doesn't completely solve the issue.

PR #1039, opened 2016-07-20
Fix browserification & minification

131:
Browserify fails to resolve the "./http-proxy/" as "./http-proxy/index.js" but as "./http-proxy.js" (so nothing works)
Beeing explicit here does not cost much for http-proxy, yet it's intrinsically complicated for browserify to fix (as trailing slash might be used as a pollyfill shim for native/non-natives addons i.e.  require('url/') vs require('url') )

### DIFF
--- a/lib/http-proxy.js
+++ b/lib/http-proxy.js
@@ -1,7 +1,7 @@
 var http      = require('http'),
     https     = require('https'),
     url       = require('url'),
-    httpProxy = require('./http-proxy/index.js');
+    httpProxy = require('./http-proxy/index.js'); //use explicit /index.js to help browserify negociation in require '/lib/http-proxy' (!)
 
 /**
  * Export the proxy "Server" as the main export.
--- a/lib/http-proxy.js
+++ b/lib/http-proxy.js
@@ -1,7 +1,7 @@
 var http      = require('http'),
     https     = require('https'),
     url       = require('url'),
-    httpProxy = require('./http-proxy/');
+    httpProxy = require('./http-proxy/index.js');
 
 /**
  * Export the proxy "Server" as the main export.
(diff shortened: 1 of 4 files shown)


131:
Rebased

lib/http-proxy.js
```diff
@@ -1,7 +1,7 @@
 var http      = require('http'),
     https     = require('https'),
     url       = require('url'),
-    httpProxy = require('./http-proxy/');
+    httpProxy = require('./http-proxy/index.js');
```
jcrugzz: Could we just add a comment on why this has to happen? Otherwise it may be removed in the future without context

131:
You got it :cake:

lib/http-proxy.js
twhid: This also breaks other packages depending on [node-resolve](https://github.com/substack/node-resolve/issues/51) (including Jest).

I'd be happy to submit another PR isolated to just this change. Let me know.

131:
Last two commits are also related to code tooling, when trying to minify the code (through uglify-js), AST fail to detect function scope (& usage) when what's intended here was really simple (just register a list of "named based hooks"). So switching to a plain old vanilla object is working fine and keep the code sane.

sorenvind:
What is the status of this? Sadly this breaks compatibility with Jest :/

131:
@sorenvind , are you saying my patch breaks some other module ? (or are you waiting for this to be merged ?)

jcrugzz:
Will get this merged in today. Sorry for the wait.

131:
With the change you requested :+1:

jcrugzz:
Cherry picked. Thanks!

twhid:
:-) thx!

sorenvind:
Wow guys! That was so much faster than I expected! @131 I talked about this being needed to fix the current lack of jest compatibility. 

Thanks!

jcrugzz:
@twhid @sorenvind no problem. I would love to see the real fix in `node-resolve` but it seems that package needs some love.

PR #1050, opened 2016-08-17
Made it not to crash with omited Host http header

alexindigo:
When tools like HAProxy run health checks they don't supply http Host header,
which makes http-proxy to crash, pretty unfortunate side-effect for a health check. :)


jcrugzz:
Thanks!

alexindigo:
Thank you. Can you publish it on npm as well? :)

jcrugzz:
Done. new version at `1.15.0`

alexindigo:
Thanks a lot.

PR #1051, opened 2016-08-17
Fix for Reason-Phrase being overwritten on proxy response.

cchamberlain:
Calling res.writeHead has the side effect of setting the Reason-Phrase back to node defaults regardless of what is in the response status line.  I'm using Reason-Phrase codes to sub-route api responses and they were all being reset.  This change only sets the statusMessage (Reason-Phrase) if it exists on the proxyRes and is successfully passing it through in my tests.

from [docs](https://nodejs.org/api/http.html#http_response_statusmessage):

image

So if just calling writeHead and not passing the optional statusMessage second argument, I believe it is getting reset to undefined and then defaulting it. I'm not sure if there is a good reason for using writeHead in general here since it seems like setting statusCode / statusMessage properties directly accomplishes the same with less side effects, could definitely be overlooking something.

What I see before change:
image

What I see after change:
image

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -108,7 +108,9 @@ var redirectRegex = /^201|30(1|2|7|8)$/;
    * @api private
    */
   function writeStatusCode(req, res, proxyRes) {
-    res.writeHead(proxyRes.statusCode);
+    res.statusCode = proxyRes.statusCode;
+    if(proxyRes.statusMessage)
+        res.statusMessage = proxyRes.statusMessage;
   }
 
 ] // <--
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -109,8 +109,9 @@ var redirectRegex = /^201|30(1|2|7|8)$/;
    */
   function writeStatusCode(req, res, proxyRes) {
     res.statusCode = proxyRes.statusCode;
-    if(proxyRes.statusMessage)
-        res.statusMessage = proxyRes.statusMessage;
+    if(proxyRes.statusMessage) {
+      res.statusMessage = proxyRes.statusMessage;
+    }
   }
 
 ] // <--


lib/http-proxy/passes/web-outgoing.js
```diff
@@ -108,7 +108,9 @@ var redirectRegex = /^201|30(1|2|7|8)$/;
    * @api private
    */
   function writeStatusCode(req, res, proxyRes) {
-    res.writeHead(proxyRes.statusCode);
+    res.statusCode = proxyRes.statusCode;
+    if(proxyRes.statusMessage)
+        res.statusMessage = proxyRes.statusMessage;
```
jcrugzz: 2 space tabs

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -108,7 +108,9 @@ var redirectRegex = /^201|30(1|2|7|8)$/;
    * @api private
    */
   function writeStatusCode(req, res, proxyRes) {
-    res.writeHead(proxyRes.statusCode);
+    res.statusCode = proxyRes.statusCode;
+    if(proxyRes.statusMessage)
```
jcrugzz: space after if, also wrap in `{ }`

cchamberlain:
Sorry for the slow response this dropped off my radar for a bit.

cchamberlain:
@jcrugzz - Any time frame on  when / if this will get merged? Not trying to rush, I just got some prod dependencies going out this week that are reliant on the status reason being passed through so I will have to fork if it does not get accepted soon.

jcrugzz:
@cchamberlain sorry it took so long. Thanks!

briman0094:
This breaks plugins such as [harmon](https://www.npmjs.com/package/harmon) that patch the `writeHead` function. Node supports `writeHead( statusCode, statusMessage, headers )`, which should be used instead of manually setting the status code and status message here.

cchamberlain:
@briman0094 - I considered that solution as well but being unaware of harmon's dependency, went with this one since it seemed lighter weight (setting the fields directly has less of this to worry about) - 

image

I can modify / open a new PR? Should we implement the `headers` parameter to allow upstream headers to be passed through or is this already implemented elsewhere?

briman0094:
@cchamberlain I don't think we need to implement `headers` because the previous version of http-proxy that called writeHead didn't implement it. There's another pass in web-outgoing.js that writes the headers from `proxyRes.headers` into the response by calling `res.setHeader(...)`. I had a PR that got merged in a few days ago which fixes the issue and it seems to be working fine.

cchamberlain:
@briman0094 - Awesome, guess my work here is done!

PR #1061, opened 2016-09-14
Properly write response header optionally including statusMessage

briman0094:
This re-adds the call to writeHead in [web-outgoing](/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-outgoing.js#L110) so that other middleware which depend on or patch the response.writeHead method (such as [harmon](/No9/harmon)) don't break.

Depends on No9/harmon#38 being accepted


jcrugzz:
@briman0094 @No9 would love a harmon test so this type of breakage could be prevented in the future. Thanks for the contribution!

PR #1062, opened 2016-09-19
Add options to override the stream destination for the proxy response

briman0094:
This adds the ability to override the stream into which the proxy response will be piped, e.g. to allow the proxy response to be piped through a transform stream before being sent back to the client.


briman0094:
Closing...accidentally grabbed the commit from my other PR with this one

PR #1073, opened 2016-09-30
Add node-http-proxy-json

langjt:
I found that there are many requirements about  to modify the response results in Issues.
Sometimes the proxy server only returns the JSON. For example, call API from the server.
Harmon is for HTML/XML, but for JSON feels some heavy. So I create this PR to help others.


indexzero:
It is 100% possible to do this with `http-proxy` today. Would prefer not to confuse / fracture the audience of users.

Glad that it works for your use case though 😸

PR #1076, opened 2016-10-04
Revert "Properly write response header optionally including statusMessage (#1061)"

lolychee:
I got the same issue: https://github.com/indexzero/http-server/issues/244#issuecomment-250935798

The reason is union only support `response.writeHead(statusCode, [headers])`
https://github.com/flatiron/union/blob/master/lib/response-stream.js#L72


jcrugzz:
@lolychee this is not the proper fix for this problem. Reverting this commit breaks other projects and the root of the cause in this issue is `union`. The fix belongs in there as the problem is with it writing bad headers.

PR #1078, opened 2016-10-05
not setting 'connection' header in case of http2

nnIIInnjA:
when using with 'http2', http2 throws an error 'can't set deprecated header : connection'


nnIIInnjA:
two test case added. 
if 'connection' header exists in req Object and when it doesn't.

Issue #930, opened 2015-12-27
Can't set headers after they are sent


xolf:
I just got a very strange error. The proxy tries to send the header twice and I have no idea why.

```
_http_outgoing.js:335
    throw new Error('Can\'t set headers after they are sent.');
          ^
Error: Can't set headers after they are sent.
    at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:335:11)
    at g:\nodejs\http-proxy\node_modules\http-proxy\lib\http-proxy\passes\web-outgoing.js:85:11
    at Array.forEach (native)
    at Array.writeHeaders (g:\nodejs\http-proxy\node_modules\http-proxy\lib\http-proxy\passes\web-outgoing.js:84:35)
    at ClientRequest.<anonymous> (g:\nodejs\http-proxy\node_modules\http-proxy\lib\http-proxy\passes\web-incoming.js:151:20)
    at ClientRequest.emit (events.js:107:17)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:419:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:111:23)
    at Socket.socketOnData (_http_client.js:310:20)
    at Socket.emit (events.js:107:17)
```

My `proxy.js` file: https://gist.github.com/xolf/f79a1f939d7dc64c9894

adambiggs:
+1 I'm having the same problem.

pyper:
This situation happens when request() call throws an error.  It looks like you have a race condition between your response error function and the web proxy.  What is happening is the request() call is failing and the custom error handler and the proxy are fighting over the same response object.  res.headersSent will not provide you protection because after both times you use them there is many asynchronous callbacks past them.   The stack trace I am getting is indicating the error is originating in the proxy which mean the header flush was likely triggered by res.end() on line 31.  This f’d up the proxy since it can no longer write headers itself because the headers were sent there.  You have to rethink the fundamental structure of this app to account for the for asynchronous nature of node and http-proxy.

jasonniebauer:
+1 I'm having the same problem

xolf:
Is there a way to ignore this error and force the server keep on running?

bubenshchykov:
+1

gabrielmancini:
+1

Issue #908, opened 2015-11-16
Running 'setHeader' during 'proxyReq' event fails for POST requests


cdl:
Right now, it looks like trying to do a `setHeader` on the passed in `proxyReq` for `POST` requests fails. (see below:)

```js
proxy.on('proxyReq', function(proxyReq, req, res, options) {
  proxyReq.setHeader('origin', 'https://example.com');
});
```

jcrugzz:
@cdl what node version does this happen in? Could you write a reproducible test case? I want to establish if this is a deterministic issue and if not what conditions FULLY create it. Thanks!

cdl:
@jcrugzz Happens in node `4.2.1`. Will see about stubbing out a reproducible test case for it for sure.

jcrugzz:
@cdl can you upgrade to `node@4.2.2` there was an `http` bug in `4.2.1` that was resolved in the patch version. See if upgrading resolves this one as well. Let me know either way! Thanks

cdl:
@jcrugzz Just tested in `node@4.2.2`, issue still persists. Will write up a gist demonstrating the issue shortly.

jcrugzz:
@cdl sounds good, and if you can encapsulate it into a test that you can PR to this project that would get us closer to a solution :). Thanks for the update!

stickystyle:
I had the same issue after upgrading to 1.12.0, downgrading back to 1.11.3 resolved it for me.

jcrugzz:
@stickystyle could you post a sample of the code that you use for `http-proxy`? This doesn't seem like a bug in `http-proxy` since its part of the `http` semantic so this is very curious

stickystyle:
@jcrugzz I've just made a scrubbed gist of the `app.js` and `package.json` that I'm using. It's a simple proxy that serves to layer JWT and caching on top of a RESTful API.
I too did a compare of the tag to master and didn't see anything that stood out, so it makes me think that one of the `http-proxy` dependencies that got updated is triggering it.

https://gist.github.com/stickystyle/a35fa8bc7461018aad83

stickystyle:
For me, this issue has gone away in the most recent v1.13.2

Issue #889, opened 2015-10-03
proxyReq.setHeader doesn't seem to work


caske33:
The example 

```proxy.on('proxyReq', function(proxyReq, req, res, options) {
  proxyReq.setHeader('X-Special-Proxy-Header', 'foobar');
});```

from the readme doesn't seem to work. It just uses the header from the req (or leaves it out if the request header was empty).

tanaydin:
Hi there,

It works with my code. There should be something else. I'm using it like..


        proxy.on('proxyReq', function proxyRequestReceiver(proxyRequest, request) {
          var chunks = [];
          request.on('data', function proxyRequestData(chunk) {
            chunks.push(chunk);
          });
          request.on('error', function proxyRequestError() {
            console.log('request on error');
          });
          proxyRequest.setHeader('host', self.host.hostname);
          ...
          ...
        });

caske33:
does this also work with custom headers? And should it be all lower cases? Or should the example in the [README](https://github.com/nodejitsu/node-http-proxy/blob/master/README.md#setup-a-stand-alone-proxy-server-with-proxy-request-header-re-writing) work like it's explained there?

jcrugzz:
In node, headers are automatically lowercased.
https://nodejs.org/api/http.html#http_http

On Sat, Oct 31, 2015 at 10:00 AM, Jens Claes <notifications@github.com>
wrote:

> does this also work with custom headers? And should it be all lower cases?
> Or should the example in the README
> <https://github.com/nodejitsu/node-http-proxy/blob/master/README.md#setup-a-stand-alone-proxy-server-with-proxy-request-header-re-writing>
> work like it's explained there?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nodejitsu/node-http-proxy/issues/889#issuecomment-152738582>
> .
>



-- 
Jarrett Cruger
Software Engineer at GoDaddy <https://www.godaddy.com/> Inc.

caske33:
But would it work if you override headers with caps, as is done in the readme and in the description of my issue?

uicoded:
Got to the same point. Blindly trusted README

mpietruschka:
@entropitor Does this bug still exist? Meanwhile, do you know why you header hasn't been set?

Issue #867, opened 2015-08-21
Avoid error Uncaught Error: Can't set headers after they are sent.


abarre:
In this portion of the code : https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L148...L157, there is no verification if the headers of the response have already been sent to the client.
It happens in my case when an error is fired and after I return an error 500 to the client.

Fix :
This portion of code can be wrapped by `if(!res.finished){}` to prevent the exception `Uncaught Error: Can't set headers after they are sent.`.

jcrugzz:
@abarre could you show me the full code where this happens? This shouldn't happen to you as the consumer but I'm guessing you are writing your own headers in a non error case?

abarre:
My error handler is something similar to the example below : 

```js
proxy.on('error', function (err, req, res, target) {
  if (err.message.indexOf("ETIMEDOUT") !== -1) {
    res.writeHead(globals.TIMEOUT_ERROR, "socket hang up);
    res.end();
  }
 else.... (for other error case)
};
```

A response is returned to the client in the error handler.

jcrugzz:
@abarre so this happens in a case where both `response` and `error` events get emit? Does it throw specifically on `res.writeHead(globals.TIMEOUT_ERROR, "socket hang up);`?

jcrugzz:
If so your fix seems like it would be correct. I just want to establish why this is happening exactly :).

abarre:
Hum, I don't remember well, I think that an `error` event is also emitted when we call the proxyReq.abort method.

giuliopaci:
I am also experiencing very similar issue, but on lines 112-115.
In my case I am trying to remove cookie header from the request, before submitting the request to the proxyed server. The code is working with GET requests, but not with POST. According to "http" manual, this is expected as with GET it is not usual that headers are sent before end() function is called, while with POST requests, headers are sent as soon as the first data are sent.
Moreover my understanding is that the listener code and the rest of the code are executed asynchronously, so there is no success guarantee with GET method as well, but it is just much more probable to succeed. Am I right?

In my specific case headers are sent immediately during http.request creation. In this case it would be more useful to add an event (hopefully a synchronous one) at the end of common.setupOutgoing(), so that it would be possible to modify options instead of the request.

bubenshchykov:
having smth similar
the difference is that my host app is doing nothing, no additional handling
and error comes from proxied app
```
[...]

PR #1091, opened 2016-11-14
add beforeProxyRequest callback config option

rick-kilgore:
Fix for https://github.com/nodejitsu/node-http-proxy/issues/867, and possibly these others: https://github.com/nodejitsu/node-http-proxy/issues/908, https://github.com/nodejitsu/node-http-proxy/issues/930, https://github.com/nodejitsu/node-http-proxy/issues/889, https://github.com/nodejitsu/node-http-proxy/issues/680.

This solution was suggested by @giuliopaci in https://github.com/nodejitsu/node-http-proxy/issues/867.  The problem with the header re-writing example in the README is that when you do a POST, the headers are sent before the 'proxyReq' event is fired.  So you really have to add the headers before calling request().  This patch allows the user to register a synchronous callback that gets called in setupOutgoing() in which the caller can setup headers, or alter the outgoing request() options in whatever way desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,21 +162,24 @@ modifies the outgoing proxy request by adding a special header.
 var http = require('http'),
     httpProxy = require('http-proxy');
 
+// To modify the proxy connection before data is sent, you can supply a
+// 'beforeProxyRequest' handler function in the passed-in config options.
+// Before the request is proxied to the target host, http-proxy will call
+// this handler with the following arguments:
+// (http.IncomingMessage req, Object options, Object requestOutgoingOptions).
+// This mechanism is useful when you need to modify the proxy request before
+// the proxy connection is made to the target.
+//
+function beforeProxyRequest(req, options, outgoingOptions) {
+  outgoingOptions.headers = outgoingOptions.headers || { };
+  outgoingOptions.headers['X-Special-Proxy-Header'] = 'foobar';
+});
+
 //
 // Create a proxy server with custom application logic
 //
-var proxy = httpProxy.createProxyServer({});
-
-// To modify the proxy connection before data is sent, you can listen
-// for the 'proxyReq' event. When the event is fired, you will receive
-// the following arguments:
-// (http.ClientRequest proxyReq, http.IncomingMessage req,
-//  http.ServerResponse res, Object options). This mechanism is useful when
-// you need to modify the proxy request before the proxy connection
-// is made to the target.
-//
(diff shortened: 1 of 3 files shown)


Jeremy-iCars:
👍

jcrugzz:
@Jeremy-iCars @rick-kilgore in this case why not just pass the headers in? I guess we could be more defensive in this case and not fire the event on post requests if it can have this type of unexpected consequences. I just think adding more options isnt the right solution.

rick-kilgore:
@jcrugzz Pass the headers in where?  In my case, I need to dynamically choose what to send in the way of headers based on the request.  Specifically, I'm supplying a default header value only if the original request does not contain the header.

jcrugzz:
@rick-kilgore before you call `proxy.web` with the `req` and `res` you can optionally pass in headers as options. You can see it in the docs [here](https://github.com/nodejitsu/node-http-proxy#options) and in the code [here](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/common.js#L46-L48).

giuliopaci:
My current understanding, is that the "proxyReq" callback is called too late for the purpouse of changing headers. What about moving the event emission of "proxyReq" at the end of setupOngoing, without adding a new event?

As far as I understand from node-https-proxy, the only purpouse for this callback is to change headers, and this do not require a socket.

In my case I want to remove headers and, sometimes, adding new headers on the basis of the reveived ones, so the flexibility provided by proxyReq callback seems to be required.

rick-kilgore:
@jcrugzz - My use case is that the incoming request may already have a header.  I want my proxy to set a particular header to some default value only if the incoming request does not have that header.  I can't just pass in the header and default value at setup time, because for incoming requests that already have that header, I don't want to set it to the default value.  I want to keep the value that the incoming request already has.

rick-kilgore:
@jcrugzz - now I see what you mean.  I somehow got the mistaken impression that the function I pass into http.createServer() was being called only once at startup, instead of once for every request.

I think my confusion was partly enabled by the example for setting headers, though.  You might want to change the example in the README to do it the way you're suggesting, since it won't work for POST requests the way it is.

aleclerc:
What if the header to be set is conditional?

For instance, I have a post request that should be "If this header `foo` is set on the incoming request, then also add header `bar`

@jcrugzz Is this currently supported?

lrettig:
+1, how do I do conditional header changes? Thanks.

Issue #1029, opened 2016-06-26
Response field key should use raw key name


fundon:
See https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-outgoing.js#L83-L89.

The `proxyRes.headers` are in lowercase. But the raw headers is not.
So I thought that we should use `proxyRes.rawHeaders` instead of `proxyRes.headers`.

fragsalat:
:+1:

oleynikd:
👍

jcrugzz:
Fixed in 1.16.0

pachirel:
I added `preserveHeaderKeyCase` option. If the option is enabled, you can get raw key name. #1104

PR #1098, opened 2016-12-01
Keep original letter case of response header keys

pachirel:
This PR fixes the issue https://github.com/nodejitsu/node-http-proxy/issues/1029

In RFC, HTTP header key is case-insensitive (ref: https://github.com/nodejs/node-v0.x-archive/issues/1954).
But, proxy server should care not to change the original request unintentionally ✨

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -83,32 +83,32 @@ module.exports = { // <--
    * @api private
    */
   writeHeaders: function writeHeaders(req, res, proxyRes, options) {
-    var rewriteCookieDomainConfig = options.cookieDomainRewrite;
+    var rewriteCookieDomainConfig = options.cookieDomainRewrite,
+        setHeader = function(key, header) {
+          if (header != undefined) {
+            if (rewriteCookieDomainConfig && key.toLowerCase() === 'set-cookie') {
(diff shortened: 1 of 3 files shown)


lib/http-proxy/passes/web-outgoing.js
```diff
@@ -87,15 +87,30 @@ module.exports = { // <--
...
-      if (header != undefined) {
-        if (rewriteCookieDomainConfig && key.toLowerCase() === 'set-cookie') {
-          header = common.rewriteCookieDomain(header, rewriteCookieDomainConfig);
+    // message.rawHeaders is added in: v0.11.6
+    // https://nodejs.org/api/http.html#http_message_rawheaders
+    if (proxyRes.rawHeaders != undefined) {
+      for (var i = 0; i < proxyRes.rawHeaders.length; i += 2) {
+        var key = proxyRes.rawHeaders[i];
+        var header = proxyRes.rawHeaders[i + 1];
+        if (header != undefined) {
```
jcrugzz: can we put this duplicate logic in a function we call with `key` and `header`? 

```js
function setHeader(key, header) {}
```

pachirel:
@jcrugzz Thank you for your quick response 😄  I fixed.

jcrugzz:
thanks!

PR #1099, opened 2016-12-01
Handle errors for forward request

maartenth:
Errors in forward requests were not handled. This caused my process to exit after an `ECONNREFUSED`.

Errors in target request and the original request were already properly handled. I added forward request error handling the same way.

Added a test case proving the fix.


jcrugzz:
LGTM 👍. Thanks

PR #1100, opened 2016-12-02
Fix newly introduced error in handler for forward ECONNREFUSED

maartenth:
In my merged PR #1099, I was too fast. I didn't make enough changes in the error handler and for `ECONNREFUSED` this introduced a new error, since `proxyReq` was not defined.

To fix it, I wrapped the error handler in a `createErrorHandler()` function to specify which request should be handled (`proxyReq` or `forwardReq`).


jcrugzz:
@maartenth We are both at fault here as I missed it too ;). Thanks for contributing! Please let me know if the forward proxy is working as expected as im not sure if it was originally written to spec.

Issue #1101, opened 2016-12-03
Set-Cookie headers do not get combined in 1.16


tomlagier:
Minor version bump of http-proxy caused our Ember-cli app to only forward a single cookie from the back-end, rather than all cookies. I believe this is because of the rewrite to web-outgoing.js/writeHeaders.

```
 Set-Cookie: GOP_LOGIN=1001; Path=/; Domain=localhost
 Set-Cookie: GOP_ACCOUNT=1001; Path=/; Domain=localhost
 Set-Cookie: PLAY_SESSION="xxx"; Path=/
```

Gets turned into this in the response header:

```
Set-Cookie:GOP_ACCOUNT=1001; Path=/; Domain=localhost
```

kriswill:
@jcrugzz @tomlagier this should be fixed now, please retest!

jcrugzz:
Yes this should be fixed with the changes made by @kriswill. Will reopen if this is still a problem. My apologies for the trouble.

PR #1102, opened 2016-12-04
Enable proxy response to have multiple Set-Cookie raw headers #1101

kriswill:
Implementation of proxy HTTP raw headers response processing in 1.16.0 fails to account for multiple `Set-Cookie` headers.  This patch accounts for multiple Set-Cookie header, and sets the source response headers to an array.

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -84,12 +84,19 @@ module.exports = { // <--
    */
   writeHeaders: function writeHeaders(req, res, proxyRes, options) {
     var rewriteCookieDomainConfig = options.cookieDomainRewrite,
+        // In proxyRes.rawHeaders Set-Cookie headers are sparse.
+        // so, we'll collect Set-Cookie headers, and set them in the response as an array.
+        set_cookies = [],
         setHeader = function(key, header) {
           if (header != undefined) {
-            if (rewriteCookieDomainConfig && key.toLowerCase() === 'set-cookie') {
-              header = common.rewriteCookieDomain(header, rewriteCookieDomainConfig);
+            if (key === 'Set-Cookie' || key === 'set-cookie') {
+              if (rewriteCookieDomainConfig) {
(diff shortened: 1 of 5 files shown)


lib/http-proxy/passes/web-outgoing.js
```diff
@@ -99,18 +106,21 @@ module.exports = { // <--
 
     // message.rawHeaders is added in: v0.11.6
     // https://nodejs.org/api/http.html#http_message_rawheaders
-    if (proxyRes.rawHeaders != undefined) {
+    if (proxyRes.rawHeaders !== undefined) {
```
jcrugzz: The current code here supports if it is `undefined` or `null`, we should be sure that it is strictly undefined.

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -84,12 +84,19 @@ module.exports = { // <--
    */
   writeHeaders: function writeHeaders(req, res, proxyRes, options) {
     var rewriteCookieDomainConfig = options.cookieDomainRewrite,
+        // In proxyRes.rawHeaders Set-Cookie headers are sparse.
+        // so, we'll collect Set-Cookie headers, and set them in the response as an array.
+        set_cookies = [],
```
jcrugzz: camelCase

lib/http-proxy/passes/web-outgoing.js
```diff
@@ -84,12 +84,19 @@ module.exports = { // <--
...
         setHeader = function(key, header) {
           if (header != undefined) {
-            if (rewriteCookieDomainConfig && key.toLowerCase() === 'set-cookie') {
-              header = common.rewriteCookieDomain(header, rewriteCookieDomainConfig);
+            if (key === 'Set-Cookie' || key === 'set-cookie') {
```
jcrugzz: we can still do `key.toLowerCase() === 'set-cookie'`

jcrugzz:
cherry-picked, thanks!

pachirel:
@kriswill Thanks to fix this issue. I apologize to any inconvenience.

Issue #1103, opened 2016-12-05
autoRewrite is broken in 1.16


apalchys:
It seems the following commit introduces a bug and "autoRewite" doesn't work properly now:  https://github.com/nodejitsu/node-http-proxy/commit/4edbb62cc5881b20e7dae5e6240e693b03fd3609

Method [setRedirectHostRewrite ](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-outgoing.js#L71) changes "location" header, but [writeHeaders ](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-outgoing.js#L107) ignores it and uses `rawHeaders` instead of `headers`

jcrugzz:
@apalchys could you give some more information on whats not working? what changed in the expectations? I apologize if all the conditions were not covered but the same logic is being run but is the value of the header not what is expected?

apalchys:
@jcrugzz you can find more details in a PR opened by @pachirel https://github.com/nodejitsu/node-http-proxy/pull/1104

pachirel:
I have not tried to reproduce this issue yet.
Any additional information are welcome. :)
2016年12月6日(火) 19:40 Andrei <notifications@github.com>:

apalchys:
@pachirel  your PR looks good and should fix autoRewrite because it doesn't overwrite header value. thanks.

pachirel:
@apalchys 1.16.2 is available. please update and check your app. thanks to @jcrugzz.

myfancypants:
I don't know if this would be the right place for feedback but in lib/http-proxy/passes/web-outgoing.js
```
res.setHeader('Set-Cookie', setCookies.length === 1 ? setCookies[0] : setCookies);
```

from 1.16.1 prior to the roll back was a breaking change. Our tooling was expecting a consistent array of cookies returned. The return of a string or an array was causing everything to break. Working locally and simply returning

```
res.setHeader('Set-Cookie', setCookies);
```

got us back up and operational.

I don't know if this impacts autoRewrite specifically, but figured I could provide some feedback on my own investigations into the matter

apalchys:
@pachirel it works now. thanks a lot!

PR #1104, opened 2016-12-06
[WIP] Revert default behavior of writeHeaders method

pachirel:
My change (#1098) has side effect because of the header values of `rawHeaders` are not always same with the values of `headers`. ( #1101, #1102, #1103) 🙇 

This PR includes these changes to solve it:

1. Overwrite only key, not value.
2. Overwrite key if raw header key is available.
3. Add `preserveHeaderKeyCase` option (default: false).

@jcrugzz  I'd like to hear your thought. 🍣 

I have to add specs before marge this.
Sorry for any inconvenience, and thanks to all.


jcrugzz:
@pachirel ill push this out now, submit a follow up PR with any additional changes you wanted to make so we can get this fixed.

pachirel:
Sure :+1:
2016年12月6日(火) 23:51 Jarrett Cruger <notifications@github.com>:

Issue #1107, opened 2016-12-13
preserveHeaderKeyCase not working for web incoming


mixxen:
The option ```preserveHeaderKeyCase``` works for web outgoing only

mixxen:
@pachirel I think you are the author of ```preserveHeaderKeyCase```. Could you take a look at this issue and pull request?

pachirel:
@mixxen thanks for your PR! I left my comments in the PR 👍

PR #1108, opened 2016-12-13
Preserve header case for incoming requests.

mixxen:
#1107

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -25,7 +25,7 @@ common.isSSL = isSSL;
  * @param {Object} Options Config object passed to the proxy
  * @param {ClientRequest} Req Request Object
  * @param {String} Forward String to select forward or target
- * 
+ * 
  * @return {Object} Outgoing Object with all required properties set
  *
  * @api private
@@ -41,7 +41,24 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
   );
 
   outgoing.method = req.method;
-  outgoing.headers = extend({}, req.headers);
+
+  var connectionHeaderName = 'connection';
+  var hostHeaderName = 'host';
+  if (options.preserveHeaderKeyCase && req.rawHeaders != undefined) {
+    var rawHeaders = {};
+    for (var i = 0; i < req.rawHeaders.length - 1; i += 2) {
+      var key = req.rawHeaders[i];
+      rawHeaders[key] = req.rawHeaders[i + 1];
+
+      if(key.toLowerCase() === connectionHeaderName)
+        connectionHeaderName = key;
+      else if(key.toLowerCase() === hostHeaderName)
+        hostHeaderName = key;
+    }
+    outgoing.headers = extend({}, rawHeaders);
+  }  else {
+    outgoing.headers = extend({}, req.headers);
+  }
 
   if (options.headers){
     extend(outgoing.headers, options.headers);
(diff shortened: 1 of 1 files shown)


lib/http-proxy/common.js
```diff
@@ -41,7 +41,24 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
...
   outgoing.method = req.method;
-  outgoing.headers = extend({}, req.headers);
+
+  var connectionHeaderName = 'connection';
+  var hostHeaderName = 'host';
+  if (options.preserveHeaderKeyCase && req.rawHeaders != undefined) {
+    var rawHeaders = {};
+    for (var i = 0; i < req.rawHeaders.length - 1; i += 2) {
+      var key = req.rawHeaders[i];
+      rawHeaders[key] = req.rawHeaders[i + 1];
```
pachirel: Learning from #1104, I prefer to use `req.headers[key]` instead.
I don't check the case of incoming, though.

pachirel:
I'm welcome this PR, because I didn't care about incoming request.
But I'm not a maintainer, please take their review ;)

IMO, Existing specs seems not to cover additional behavior. Would you please add specs?

honzajavorek:
I think we've just bumped into some problems related to this. Is there a way we can help to get this merged and released?

jarrodek:
It's worth to mention that the http message spec describes headers as a case insensitive. Therefore clients/servers that relays on headers cases sensitivity are  not complying with the spec and should not be expected to work properly.
Developers shouldn't get used to bad practices (especially if something is against the spec) and therefore this PR shouldn't be accepted.
That's only my opinion.

mixxen:
@jarrodek The issue here is dealing with 3rd party software or legacy software that does not adhere to the spec and we have no control over.

honzajavorek:
The same with us. I think node-http-proxy preserving case of header names would be something similar to "be benevolent in what you accept and strict about what you produce". I totally agree it's ridiculous some software is HTTP header name case-sensitive, but software is made by humans and humans make mistakes.

This is very simple change node-http-proxy can do and it even actually makes sort of sense - IMHO the proxy should pass things "as is" to be as unnoticeable as possible, unless it really needs to touch them and modify them for some purpose.

jarrodek:
I totally agree with @honzajavorek The proxy should be totally transparent for passing data and not alter them. In this case, however, it doesn't actually changes the data. If anything, that's the problem of Node's Request object and not the proxy. This PR actually introduce a behavior where data are altered by the proxy. This PR should be sent to Node because that's the source of the problem and not this module.

mrichmond:
Just ran into this problem where for whatever reason the service I'm calling expects a header to be Mixed Case ('X-HTTP-Method-Override'). Even with the `preserveHeaderKeyCase: true` it looks like the header is being rewritten to ('x-http-method-override'). I'm able to work around this by rewriting the headers... but it's a pretty lame solution since I'd have to do it for each header and I may not now what the original case was. Why wouldn't it just leave the header casing alone in the first place??

`
const onProxyReq = (proxyReq, req) => {
  if (req.headers['x-http-method-override']) {
    proxyReq.setHeader('X-HTTP-Method-Override', req.headers['x-http-method-override']);
  }
};
`

ItsEcholot:
Just ran into this same issue as @mrichmond described... Why are headers rewritten to lower-case?

PR #1109, opened 2016-12-13
Update eventemitter3 to version 2.0.x

lpinca:
`node-http-proxy` is not affected by the only breaking change in version [2.0.0](https://github.com/primus/eventemitter3/releases/tag/2.0.0) so I think it makes sense to update as there are some nice performance improvements.


jcrugzz:
Thanks @lpinca

PR #1115, opened 2016-12-20
Update README.md: Inform users only to use target or forward, not both

chrisnew:
added hint to inform users not use use both target and forward


indexzero:
Agree that it's not as common, but it is prudent / necessary to use both of these in conjunction in several scenarios.

PR #1121, opened 2016-12-25
Improved formatting for some lines in README.md

ivannieto:
Corresponding to the explanations preceded by the '†' symbol. I had to re-read those lines when reading the doc, I think makes more reading-sense now.

Please review. Does not change the README itself.


ivannieto:
Forgot to pull before. I should remember Chuck's words... 😄 

Fixing now.

PR #1127, opened 2017-01-03
bugfix: handle toProxy option correctly

BernhardRode:
Handle toProxy option properly and get proxy settings from environment.

There is an option toProxy (true/false) for http-proxy. But it is never used afaiks.
This fix checks the option and uses the http/https proxy set in the environment variables.

Usage:
```
{
  "/api": {
    "target": "http://upstreamserver.net",
    "toProxy": true
  }
}
```

Sorry, for I had no time for unit tests (just needed a local hotfix ;) )... If you like the solution I could into this PR in more detail during the weekend.

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -50,7 +50,7 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
   if (options.auth) {
     outgoing.auth = options.auth;
   }
-  
+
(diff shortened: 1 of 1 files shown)


BernhardRode:
Created a feature brand for this PR.
See https://github.com/nodejitsu/node-http-proxy/pull/1128

PR #1128, opened 2017-01-03
bugfix: use upstream proxy

BernhardRode:
Handle toProxy option properly and get proxy settings from environment.

There is an option toProxy (true/false) for http-proxy. But it is never used afaiks.
This fix checks the option and uses the http/https proxy set in the environment variables.

Usage:
```
{
  "/api": {
    "target": "http://upstreamserver.net",
    "toProxy": true
  }
}
```

Sorry, for I had no time for unit tests (just needed a local hotfix ;) )... If you like the solution I could into this PR in more detail during the weekend.

### DIFF
--- a/lib/http-proxy/common.js
+++ b/lib/http-proxy/common.js
@@ -50,7 +50,7 @@ common.setupOutgoing = function(outgoing, options, req, forward) {
   if (options.auth) {
     outgoing.auth = options.auth;
   }
-  
+
   if (options.ca) {
       outgoing.ca = options.ca;
   }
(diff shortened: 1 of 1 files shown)


assiomatica:
Hi,

I have tested your patch, as is it's not working, below the correct code:

`if (options.toProxy) {
    var upstreamProxyServer;
    if (typeof options.toProxy === 'boolean') {
      upstreamProxyServer = url.parse(
        isSSL.test(options[forward || 'target'].protocol) ? process.env['https_proxy'] || '' : process.env['http_proxy'] || ''
      );
    } else {
      upstreamProxyServer = url.parse(options.toProxy);
    }
    
    if (upstreamProxyServer) {
      outgoing.path = options[forward || 'target'].protocol + '//' + outgoing.host + outgoing.path;
      outgoing.host = upstreamProxyServer.host;
      outgoing.hostname = upstreamProxyServer.hostname;
      outgoing.port = upstreamProxyServer.port;
    }
  }
`

PR #1130, opened 2017-01-09
Update options.changeOrigin description in docs

asvae:
changeOrigin has nothing to do with `origin` header which is confusing and harmful when working with CORS. As CORS middleware sometimes makes decisions depending on `origin` header.

`changeOrigin` option handles solely `host` header.


patrickallaert:
Correct, but in my opinion, the `changeOrigin` config should just be removed and default to `true`. See #1132 for more info.

johnhunter:
While discussion on removing changeOrigin continues, can we not update the docs to clarify what the setting actually does?

It does not change the origin header but consuming projects assume it does.

Issue #1143, opened 2017-02-27
Server listen callback with https


barto-:
I would like a callback when the server is listening. This is possible with ```http``` but it doesn't work with ```https```.

Example:

```js
// ...
var proxyServer = httpProxy.createServer({
  ssl: {
    key: fs.readFileSync('valid-ssl-key.pem', 'utf8'),
    cert: fs.readFileSync('valid-ssl-cert.pem', 'utf8')
  },
  target: 'https://localhost:9010',
  secure: true // Depends on your needs, could be false.
});

proxyServer.listen(443, function listening(err){
  console.log('Server listening');
});

```

PR #1144, opened 2017-02-27
Added server listen callback

barto-:
Added support for server listen callback #1143 .


YoannGalentin:
Is it possible to accept this pull request, please ? Or if it is not possible, Why 😞 ?
Thanks 😄

remino:
Is this project abandoned? I was looking for why callbacks weren't supported in `listen()` and found this, still open after two years.

With the last updating being on May 18, I believe we're better off not using http-proxy at all.

PR #1147, opened 2017-03-04
Fix docs for rewrite options - 201 also handled

smh:
Updates options documentation for location rewrite to include 201 responses. See #1024


indexzero:
Thanks – sorry for the long merge delay!

Issue #1113, opened 2016-12-19
Specify incoming request socket timeout


phillipj:
Hi!

I've found an issue allowing long running proxied connections to stay open. In this case, it's something we've seen related to the [webpack dev server](https://github.com/webpack/webpack-dev-server) which has this http-proxy package as its dependency via [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware).

In practise what we've seen is that long running proxied requests are being closed after 2 minutes, even though we specify a longer timeout with the **proxyTimeout** option. That in turn may make some browser automatically retry the request which failed, which is absolutely not wanted in some scenarios -- the latter isn't really this package's fault, so feel free to ignore that part.

After [digging for a while](https://github.com/webpack/webpack-dev-server/issues/369#issuecomment-266251977) I found the root cause to be in this package's request handler ([lib/http-proxy/index.js](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/index.js#L31)). By not setting a timeout on the incoming HTTP socket with [request.setTimeout()](https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_message_settimeout_msecs_callback), Node.js core sets 2 minutes by default.

Ideally we'd like to be able to set **proxyTimeout** to whatever number we want and be sure that's respected. That possibly means using the same value in `request.setTimeout()`.

Does this sound okey to you? Any other thoughts?

Refs https://github.com/webpack/webpack-dev-server/issues/369#issuecomment-266251977

/cc @Martin-Wegner

benmosher:
FWIW, I was able to work around this via 

```js
const LONG_PYTHON_TIMEOUT = 15 * 60 * 1000;
const pythonProxy = new httpProxy.createProxyServer({
  target: {
    host: 'localhost',
    port: pythonPort
  },
  proxyTimeout: LONG_PYTHON_TIMEOUT,
})
app.all('/api/*', (req, res) => 
  pythonProxy.web(req.setTimeout(LONG_PYTHON_TIMEOUT), res))
```

I just set the request timeout in the handler lambda before passing into the proxy handler itself.

That said: right or wrong, I 100% expected it to set the request timeout for me, mostly because I didn't even know that was a thing until I ran into the timeout in a long test.

So I have no idea what arguments might exist against setting the `request` timeout inside the proxy handler, but I can't imagine why that would be desired. (as someone who has considered it for all of 10 minutes now 😅)

[...]

PR #1154, opened 2017-03-17
Set incoming requesting timeout

phillipj:
This ensures the incoming request does not timeout before the target's timeout configured with `options.proxyTimeout`.

As I'm not familiar to the code base, I struggled seeing the best entry to write a test for this. Any thoughts would be appreciated.

Fixes https://github.com/nodejitsu/node-http-proxy/issues/1113

~~**Edit:** I'm looking into the existing timeout test failing~~

### DIFF
--- a/lib/http-proxy/index.js
+++ b/lib/http-proxy/index.js
@@ -68,6 +68,11 @@ function createRightProxy(type) {
         return this.emit('error', new Error('Must provide a proper URL as target'));
       }
 
+      // ensure request don't timeout before the target's configured timeout
+      if (options.proxyTimeout) {
+        req.setTimeout(options.proxyTimeout);
+      }
+
       for(var i=0; i < passes.length; i++) {
         /**
          * Call of passes functions


phillipj:
Looking at the test that fails with these changes added, I wonder if that test is wrong.

It doesn't expect the client request being made to throw an error if an error occurs. At first glance that seems weird, cause a timeout is also an error to the client initiating the request, right?

Additionally there is two very similar tests, whereas the latter succeeds cause it expects timeouts to raise client error:
- failing: [it('should proxy the request and handle timeout error (proxyTimeout)', ..)](https://github.com/nodejitsu/node-http-proxy/blob/master/test/lib-http-proxy-passes-web-incoming-test.js#L209)
- similar: [it('should proxy the request and handle timeout error', ..)](https://github.com/nodejitsu/node-http-proxy/blob/master/test/lib-http-proxy-passes-web-incoming-test.js#L243)

indexzero:
Closing as a duplicate of #1273 – if it's not a strict duplicate . from a software design perspective this belongs in the `passes` not in the root `lib/http-proxy/index.js`

indexzero:
Thanks for helping triage this issue and contribute a proposed fix though 😸

Issue #1017, opened 2016-06-01
Determining which end is causing a ECONNRESET


archa347:
We've been seeing some ECONNRESET errors in our proxy application.  The errors are being handled  so its not causing any major issues, but we're worried there might be connectivity issues with the target server.  Is there anyway we can differentiate between ECONNRESET errors coming from the client side connection versus the target side?

PepperYan:
+1, I get request package in system, it seems to send request to localhost (original url), but not my target url

yunfan:
this issue is still open made me hopeless

Issue #1167, opened 2017-05-10
Travis run failing on node 0.10/0.12 due to unpinnned socket.io-client


edmorley:
The Travis run is failing on node 0.10/0.12 with:
```
/home/travis/build/nodejitsu/node-http-proxy/node_modules/socket.io-client/node_modules/engine.io-client/node_modules/ws/index.js:9
const WebSocket = require('./lib/WebSocket');
^^^^^
...
  1) lib/http-proxy.js #createProxyServer using the ws-incoming passes should proxy a socket.io stream:
     SyntaxError: Use of const in strict mode.
      at Server.init (node_modules/socket.io/node_modules/engine.io/lib/server.js:119:16)
      at new Server (node_modules/socket.io/node_modules/engine.io/lib/server.js:65:8)
      at Function.attach (node_modules/socket.io/node_modules/engine.io/lib/engine.io.js:123:16)
      at node_modules/socket.io/lib/index.js:258:23
      at Encoder.encode (node_modules/socket.io/node_modules/socket.io-parser/index.js:138:5)
      at Server.listen.Server.attach (node_modules/socket.io/lib/index.js:251:16)
      at new Server (node_modules/socket.io/lib/index.js:57:17)
      at Function.Server (node_modules/socket.io/lib/index.js:43:41)
      at Context.<anonymous> (test/lib-http-proxy-test.js:432:20)
```
(for example https://travis-ci.org/nodejitsu/node-http-proxy/jobs/230765815)

The package `ws` has dropped support for these versions of node: 
https://github.com/websockets/ws/issues/1077

Options are to:
* Drop support for those versions of node from node-http-proxy too (they are EOL after all)
* Pin to `ws` version 1.x. However engine.io-client 3.x explicitly requires ws 2.x, so we'd need to pin to engine.io-client 2.x (the latest of which is 2.2.1), which would mean falling back from socket.io-client 2.x to 1.7.4.

Note at the moment that node-http-proxy doesn't pin socket.io-client or many other devDependencies at all:
```json
  "devDependencies": {
    "async": "*",
    "blanket": "*",
    "coveralls": "*",
    "dox": "*",
    "expect.js": "*",
    "mocha": "*",
    "mocha-lcov-reporter": "*",
    "semver": "^5.0.3",
    "socket.io": "*",
    "socket.io-client": "*",
    "sse": "0.0.6",
    "ws": "^0.8.0"
},
```

edmorley:
[...]

PR #1172, opened 2017-05-23
Store the request object in proxy errors

rowanmanning:
This would allow someone to differentiate between client errors and
target server errors in their proxy error handling, by inspecting the
error request property.

I'm not sure if this is how you'd prefer to do this, open to alternatives.
The problem I'm trying to solve is in a proxy service I maintain – I'd
like to differentiate between client and server errors in our logging.

This is a potential solution for #1017. E.g:

```js
proxy.on('error', (error, request, response, url) => {
    // Will log either "target" or "forward"
    console.log(`Error came from ${error.request.proxyRequestType}`);
});
```

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -104,6 +104,7 @@ module.exports = {
       var forwardReq = (options.forward.protocol === 'https:' ? https : http).request(
         common.setupOutgoing(options.ssl || {}, options, req, 'forward')
       );
+      forwardReq.proxyRequestType = 'forward';
 
       // error handler (e.g. ECONNRESET, ECONNREFUSED)
       // Handle errors on incoming request as well as it makes sense to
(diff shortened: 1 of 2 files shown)


rowanmanning:
Tests are failing in Node.js 0.10.x/0.12.x, but it's unrelated to my changes. See #1167.

jcrugzz:
@rowanmanning I hesitate to add such a heavy object with circular references to the error (because this could cause issues if someone tries to log the error object potentially) but I'm not opposed to attaching metadata so that we can detect the request type and any other relevant information. Is there anything besides `type` you would want to add?

rowanmanning:
Hi @jcrugzz, your hesitance makes perfect sense. I'd be happy to have `type` metadata, and can't think of anything else off the top of my head. If you don't have time to do the work yourself then I'd also be happy to have a go myself, just might need pointing in the right direction :)

PR #1175, opened 2017-05-26
test case when port 80 is being omitted in location header

ivstas:
I have an issue when trying to proxy to target with explicitly set port **80**.
In this case port 80 in location header is being omitted, which doesn't pass the check and, as a result, not being proxied.


jcrugzz:
@ivstas based on the code this seems to be expected behavior. We could technically validate `hostname` instead of `host` but there is a test that is distinctly stating that this is expected behavior. Is there someway you can ensure both target and port are consistent rather than the location being inconsistent from the target? since port 80 is the default, omitting it from target in order to make them consistent should solve the issue but maybe im not fully understanding the use case. 

@matthauck who wrote the code may be able to speak more on this but for now Ill close as test phrasing and intentions are mismatched.

Issue #1168, opened 2017-05-12
Running 'setHeader' during 'proxyReq' event fails for POST requests


lrettig:
Hello,

I'm running into the same issue that happened in #908 which was supposed to have been solved already, but it's occurring for me now on the latest v1.16.2. Trying to do a `setHeader` on the passed in `proxyReq` for a POST requests fails with:

`Error: Can't set headers after they are sent.`

It's happening for both node 4.3.2 and 6.1.0.

To reproduce:

```javascript
proxy.on('proxyReq', function(proxyReq, req, res, options) {
  proxyReq.setHeader('origin', 'https://example.com');
});
```
Thank you.

leth:
I've found a workaround:
```javascript
proxyReq.headers = proxyReq.headers || {}
proxyReq.headers['x-my-header'] = 'value'
```

XVincentX:
Most of the times the problem is because you're subscribed to the events more than once. `proxy.once('proxyReq,…)` usually solves the problem guys.

fingerpich:
I would like to change headers and path on proxyReq like the following code
```
proxyReq.path="/abcd/abc" 
```
It works well with GET requests but it doesn't with POST requests

skleeschulte:
In my case the problem was an `Expect: 100-continue` header sent by the client. See https://github.com/http-party/node-http-proxy/issues/1219

PR #1176, opened 2017-05-26
Fix "Can't set headers after they are sent" errors

### LINKED ISSUE #930: Can't set headers after they are sent

thiagobustamante:
This PR fix "Can't set headers after they are sent" errors.
That are a lot of situations where this error can occurs. In my case, it is happening because I have others middlewares (in an expressjs application that tries to proxy requests). Some of those middlewares (like [passportjs](http://passportjs.org/), or [cors](https://www.npmjs.com/package/cors)) can run ```res.end()``` and when the proxy receive a response, it is already finished.
So, it is necessary to test if we can write on the user response when the proxy response is ready.
I think it could also fix #930, #1168, #908 and #867

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -162,16 +162,22 @@ module.exports = {
 
     proxyReq.on('response', function(proxyRes) {
       if(server) { server.emit('proxyRes', proxyRes, req, res); }
-      for(var i=0; i < web_o.length; i++) {
-        if(web_o[i](req, res, proxyRes, options)) { break; }
+      if (!res.headersSent) {
+        for(var i=0; i < web_o.length; i++) {
+          if(web_o[i](req, res, proxyRes, options)) { break; }
+        }
       }
 
       // Allow us to listen when the proxy has completed
-      proxyRes.on('end', function () {
-        server.emit('end', req, res, proxyRes);
-      });
-
-      proxyRes.pipe(res);
+      if (!res.finished) {
+        proxyRes.on('end', function () {
+          server.emit('end', req, res, proxyRes);
+        });
+        proxyRes.pipe(res);
+      }
+      else {
+          server.emit('end', req, res, proxyRes);
+      }
     });
 
     //proxyReq.end();


thiagobustamante:
Tests are failing in Node.js 0.10.x/0.12.x, but it's unrelated to my changes. See #1167.

lib/http-proxy/passes/web-incoming.js
```diff
@@ -162,16 +162,22 @@ module.exports = {
 
     proxyReq.on('response', function(proxyRes) {
       if(server) { server.emit('proxyRes', proxyRes, req, res); }
-      for(var i=0; i < web_o.length; i++) {
-        if(web_o[i](req, res, proxyRes, options)) { break; }
+      if (!res.headersSent) {
+        for(var i=0; i < web_o.length; i++) {
+          if(web_o[i](req, res, proxyRes, options)) { break; }
```
frogmella: I think this is a better fix: https://github.com/nodejitsu/node-http-proxy/pull/1182

thiagobustamante: Hi @frogmella,

I think that your your commit is ok and we should avoid the writeHead function. But if the headers are already sent, does not makes sense to call ```res.statusCode = proxyRes.statusCode;```. The response status code and headers was already sent to client, so you can not modify it anymore.

jcrugzz:
Thanks for the contribution @thiagobustamante. I will cherry-pick this into my maintenance branch that will be merged in today.

jcrugzz:
superseded by #1251

Issue #953, opened 2016-02-04
'Can't set headers after they are sent' with express


jnjnljljjsw:
my node.js version is 4.2.2

app.js:
`var admin_api = require('./routes/api');
app.use('/api', api);`

api.js
`
proxy.on('proxyReq', function(proxyReq, req, res, options) {
proxyReq.setHeader('X-Special-Proxy-Header', 'foobar');
});
router.all("/*", function(req, res){ 
  proxy.web(req, res, { target: 'http://dev.lalocal.cn:8080/api' });
});
`

error:
Error: Can't set headers after they are sent.

michaellujan:
Having the same issue under load. My set up is almost identical

tomcanham:
I ran into this too; it only seems to happen for multipart (chunked) responses. I didn't dig into it too much, but I found a workaround. When you call the proxy method, you pass in an options object -- the third parameter, which contains the target. 

Well apparently, if this object contains a 'headers' object, those headers are copied to the proxied request as it's being initialized, and the timing of this copy avoids whatever race condition is causing this failure.

From my code:

```
              options = { target: target }

              if (result.user) {
                options.headers = {}

                options.headers['x-auth-user-id'] = result.user._id
                options.headers['x-auth-user-username'] = result.user.username
                options.headers['x-auth-user-roles'] = result.user.roles
              }

              return proxy.web(req, res, options)
```

bubenshchykov:
+1, happens during bigger load

Phris:
+1

michaellujan:
not sure your exact use case but I ended up writing something that handled this better. Have used it in production for over a year handling multi millions of requests, and am now open sourcing it:

[BiFrost](https://github.com/moonbus/bifrost)

gabrielmancini:
+1

PR #1182, opened 2017-06-14
issue #953: stop using writeHead

jakefurler:
Using object.keys in web-incoming.js results in a non-deterministic ordering of keys, which means that in web-outgoing writeHead might be called before setHeader, which throws an error.

Example of a server that fails to send a response : 
```
const http = require('http');

const server = http.createServer((req, res) => {
  res.writeHead(200);
  res.setHeader('Content-Type', 'text/html');
  res.setHeader('X-Foo', 'bar');
  res.end('ok');
});

server.listen(3000);
```
This will throw `Error: Can't set headers after they are sent.` when a request is made to it. If we change it to the following, it will work:

```
const http = require('http');

const server = http.createServer((req, res) => {
  res.statusCode = 200;
  // res.writeHead(200);
  res.setHeader('Content-Type', 'text/html');
  res.setHeader('X-Foo', 'bar');
  res.end('ok');
});

server.listen(3000);
```

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -129,9 +129,10 @@ module.exports = { // <--
   writeStatusCode: function writeStatusCode(req, res, proxyRes) {
     // From Node.js docs: response.writeHead(statusCode[, statusMessage][, headers])
     if(proxyRes.statusMessage) {
-      res.writeHead(proxyRes.statusCode, proxyRes.statusMessage);
+      res.statusCode = proxyRes.statusCode;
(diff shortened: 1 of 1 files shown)


frogmella:
@jcrugzz can you please review this? It's causing issues in one of my production services, and I'm having to work around it at this point in time. I think merging it would solve a lot of issues people seem to be running into.

jcrugzz:
Thanks for the contribution @jakefurler! I cherry-picked your commit to my maintenance branch which i plan to merge later tonight.

jcrugzz:
superseded by #1251

PR #1184, opened 2017-06-20
Include websocket non-upgrade response

### LINKED ISSUE #890: digest auth for websocket proxying
Hi, I am trying to proxy a WebSocket to an endpoint which requires digest authentication.

```
 var HttpProxy = require('http-proxy'),
   http = require('http');
 var wsProxy = new HttpProxy.createProxyServer({target: 'ws://' + options.cameraIp, ws: true});
 var server = http.createServer();
 server.on('upgrade', function (req, socket, head) {
  wsProxy.ws(req, socket, head);
});
```

The first attempt gives a 401 (the digest challenge) which is never returned to the browser. How can I hook in / configure http-proxy to pass the 401 back to the browser when using websockets?

Tigge:
When the server do not accept the upgrade request for websockets the
response was previously not included and sent back. Now the proxy will
include the headers and the content from the server and the response in
these cases. Fixes #890.

### DIFF
--- a/lib/http-proxy/passes/ws-incoming.js
+++ b/lib/http-proxy/passes/ws-incoming.js
@@ -77,6 +77,24 @@ module.exports = {
    * @api private
    */
   stream : function stream(req, socket, options, head, server, clb) {
+
+    var createHttpHeader = function(line, headers) {
+      return Object.keys(headers).reduce(function (head, key) {
+        var value = headers[key];
+
+        if (!Array.isArray(value)) {
+          head.push(key + ': ' + value);
+          return head;
+        }
+
+        for (var i = 0; i < value.length; i++) {
+          head.push(key + ': ' + value[i]);
+        }
+        return head;
+      }, [line])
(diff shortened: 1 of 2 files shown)


Tigge:
To expand a bit: This will allow node-http-proxy to support some authentication schemes when required by the remote server. If this is supported depends on the browser, but Chrome and Firefox for example will happily work with digest authentication with this addition.

Even if we disregard the added benefits above it seems correct for the proxy to relay back the error messages from the server, whatever it may be.

Tigge:
@jcrugzz I think the remaining test issues are caused by socket.io using const and not by any changes I have introduced. The test would probably fail on master now as well. Any thoughts on how to proceed?

jcrugzz:
@Tigge thanks! I am cherry-picking this into my maintenance branch so that I can get this merged in later tonight in one go as I go through all the PRs.

jcrugzz:
superseded by #1251

PR #1187, opened 2017-06-25
FEAT: Bindings section allows for per host name forward / target.

krassif:
## Status
**READY**

## Migrations
NO

## Description
Single proxy may handle requests targeting multiple "hosts"/ DNS entries and then distribute the load as per the caller specified host. Both target host and target port are part of the load distribution.

For instance, consider the following distribution spec:

```
var binding = {
    "yarn.pronto.peerbelt.com": {
        target: "https://yarnpkg.com:443",
        "changeOrigin": true
    },
    "google.pronto.peerbelt.com": {
        target: "https://www.google.com:443",
        "changeOrigin": true
    },
    "*.pronto.peerbelt.com": {
        target: "https://www.microsoft.com:443/en-us/",
        "changeOrigin": true
    },
    "pronto.peerbelt.com": {
        target: {
            host: "overlayads.herokuapp.com",
            port: 443,
            protocol: "https"
        },
        changeOrigin: true
    }
};
```

Based on it, HTTP request targeting:
* ```yarn.pronto.peerbelt.com``` are getting forwarded to ```https://yarnpkg.com:443```
* ```pronto.peerbelt.com``` calls ```overlayads.herokuapp.com```
* ```something-different.pronto.peerbelt.com``` follows the wildcard rule hitting ```https://www.microsoft.com:443/en-us/``` accordingly

Anything not found in the distribution spec follows the already existing proxy rules. For instance:

```
httpProxy.createServer({
  binding: binding,
  target: {
    host: "localhost",
    port: 3030
  }
}).listen(80)
.on( "error", socketErrorHandler );
```

with incoming request targeting ```my.peerbelt.com``` is simply forwarded to ```localhost:3030```.

## Todos
- [ ] Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,7 @@ proxyServer.listen(8015);
      ```
 *  **headers**: object with extra headers to be added to target requests.
 *  **proxyTimeout**: timeout (in millis) when proxy receives no response from target
+*  **binding**: object allowing for per hostname targeting / forwarding. Each object key specifies hostname glob. When the glob is matched, the options listed under the key are used by the proxy. To merge the base option config with a partial, domain glob specific configuration, specify _merge: true at the partial config root. Intended to be used with a single (port, protocol) proxy and star certificate in case SSL is needed.
 
 **NOTE:**
 `options.ws` and `options.ssl` are optional.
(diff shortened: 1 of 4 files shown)


jcrugzz:
@krassif i think this is a good feature but it can be implemented as a higher level module that consumes `http-proxy` to do this behavior. I would suggest you write the module that will accomplish this and use `http-proxy` to do the core proxying.

Issue #1090, opened 2016-11-10
Example of proxying to a server with TLS client authentication


datpham23:
Hi,

I have use case where I'm proxying to a Apache Tomcat server w/ client authentication turn on. Does node-http-proxy support mutual tls? If so is there any documentation of that?

Regards,
Dat

datpham23:
After digging through the code I finally got POC that work, hopefully this can help someone else.

`
var http = require('http'), httpProxy = require('http-proxy'); fs = require('fs');
var https = require('https');

var options = {
  pfx: fs.readFileSync('clientCert.pfx'),
  passphrase: 'password',
  rejectUnauthorized: false
};
options.agent = new https.Agent(options);

var proxyConfig = {
  target: 'https://localhost:8080/',
  agent: options.agent
};

proxyConfig = Object.assign(proxyConfig,options);

var proxy = httpProxy.createProxyServer(proxyConfig);


var server = http.createServer(function(req, res) {
  proxy.web(req, res);
});

server.listen(5050);


`

raduserbanescu:
Hi there! Your snippet helped me in finding a solution to a similar problem.

I also made a Pull Request #1209, adding an example to the official use cases in order to help other people in the future.

Issue #826, opened 2015-05-20
Proxy with client certificates


javierpavon2000:
I'm trying to perform a proxy to an url that needs a client certificate. The context is:

* I've installed a p12 file in the browser and after that I can access to the endpoint https://aaa.com/webservice:4278
* I've extracted the key and the certificate of the p12 file installed in the browser

Now, I've defined an object like that
```
var objConf = { target: 'https://aaa.com/webservice:4278',
  secure: false,
  headers: { host: 'scytl-sessmgm.cloudapp.net:4278' },
  ssl:
   { key: '-----BEGIN RSA PRIVATE KEY----- ..... -----END RSA PRIVATE KEY-----',
     cert: '-----BEGIN CERTIFICATE----- ... -----END CERTIFICATE-----' 
	}
}
```

and the, I create the proxy
```
var proxy = httpProxy.createProxyServer(objConf);
```

But, it doesn't work, the request always returns a 403 status code.

Am I doing something wrong? Any help?
Thanks

kjlaw89:
I'm curious if there's a way to do this too. I opened a Stack Overflow question on it the other day: http://stackoverflow.com/questions/30331618/node-http-proxy-to-https-server-with-client-certificate

briangreenery:
For sure it's possible for http-proxy to use a client certificate. I just answered that Stack Overflow question, although I used the `key`, `cert`, and `ca` options instead of the `pfx` option. I think the example will still work if you use the `pfx` option instead.

Here's a working demo: https://gist.github.com/briangreenery/e2f03c0058e3589a9acf

PR #1209, opened 2017-10-16
Add use case for proxy to HTTPS using a PKCS12 client certificate

raduserbanescu:
Add an example under the [Using HTTPS](https://github.com/nodejitsu/node-http-proxy#using-https) section on how to proxy requests to an HTTPS server that requires a PKCS12 client certificate (like the one you normally import in your browser).

I recently had this exact use case and while looking for a solution I stumbled upon Issues #826 and #1090 (which helped me in finding a solution).

Browsing the changes of Pull Request #590, I saw that you can pass extra parameters in the `target` object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,24 @@ httpProxy.createServer({
 }).listen(443);
 ```
 
+##### HTTP -> HTTPS (using a PKCS12 client certificate)
+
+```js
+//
+// Create an HTTP proxy server with an HTTPS target
+//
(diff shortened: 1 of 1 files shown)


jcrugzz:
Thanks for the contribution @raduserbanescu ! Added to my maintenance branch with cherry-pick so it will be merged in later today once i get through these PRs.

jcrugzz:
superceded by #1251

Issue #1066, opened 2016-09-21
how to rewrite the http methods 


cuiyongjian:
i want to change all the http "get" request   to    the target  by "post" method.

but i can not find  the node-http-proxy documents to resolve this problem.

AydinChavez:
Created a pull request 
https://github.com/nodejitsu/node-http-proxy/pull/1231
for this.

PR #1231, opened 2018-01-23
option for overriding HTTP-header

AydinChavez:
Add method parameter to options for overriding the proxy-outgoing HTTP-method.

Simple solution for issue: https://github.com/nodejitsu/node-http-proxy/issues/1066

Example usage:

```
  // You can define here your custom logic to handle the request
  // and then proxy the request.
  proxy.web(req, res, { target: 'http://127.0.0.1:9200', method: 'GET' });
  console.log("req: " + req);
  console.log("res: " + res);
});
```
You could also wrap it into a conditional statement like:
```
if ( req.method == "POST" ) {
   proxy.web ...
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,27 @@ node-http-proxy
 websockets. It is suitable for implementing components such as reverse
 proxies and load balancers.
 
+### Fork details
+This fork offers the option for overriding the proxy-outgoing HTTP-method.
+
+Example usage:
+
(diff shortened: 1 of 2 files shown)


jcrugzz:
Ok, this doesn't seem harmful. Will add this to my maintenance branch @AydinChavez. Thanks for the contribution!

jcrugzz:
superceded by #1251

PR #1232, opened 2018-01-31
Added timeout option to docs

jlaamanen:
Hi, it seems that `timeout` option is missing in the docs ([code reference here](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L39-L53)). This would be a useful addition because I wasted several hours trying to figure out what was wrong, until I happened to find this option from the library source code.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,8 @@ proxyServer.listen(8015);
      }
      ```
 *  **headers**: object with extra headers to be added to target requests.
-*  **proxyTimeout**: timeout (in millis) when proxy receives no response from target
+*  **proxyTimeout**: timeout (in millis) for outgoing proxy requests
+*  **timeout**: timeout (in millis) for incoming requests
 
 **NOTE:**
 `options.ws` and `options.ssl` are optional.


pataltechnologies:
This piece is pretty critical if you want to set the `proxyTimeout` for anything longer than 2 minutes since the proxy HTTP server request timeout is set to 2 minutes and will cause retries every 2 minutes. If anybody is uneasy about using the undocumented `timeout` property in this library you can also manually set the timeout on the HTTP server after you create it as well.

Here is a modified snippet from the "[Setup a stand-alone proxy server with custom server logic](https://github.com/nodejitsu/node-http-proxy#setup-a-stand-alone-proxy-server-with-custom-server-logic)" section of the README:

```js
  proxy.web(req, res, { target: 'http://127.0.0.1:5060' });
});
server.setTimeout(1000 * 60 * 60);
server.listen(5050);
```

jcrugzz:
just going to cherry-pick this into my maintenance branch, but thanks for the contribution @jlaamanen

jcrugzz:
superceded by #1251

PR #1250, opened 2018-04-17
Update common.js

zjgnlzq:
repair bug，when use webpack-dev-server proxy


jcrugzz:
@zjgnlzq Thanks for the contribution! this behavior is  currently done only when `changeOrigin` option is set to true. This is not current default behavior and would be a breaking change that I'm not willing to make at this time. I would check out the `changeOrigin` option and set it to true in your use case (webpack-dev-server) https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/common.js#L100-L105

PR #1251, opened 2018-04-19
Maintenance

jcrugzz:
Running branch for merging PRs into that will properly run tests and pins dependencies as it should be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,24 @@ proxyServer.listen(8015);
 *  **proxyTimeout**: timeout (in millis) for outgoing proxy requests
 *  **timeout**: timeout (in millis) for incoming requests
 *  **selfHandleRequest** true/false, if set to true, none of the webOutgoing passes are called and its your responsibility ro appropriately return the response by listening and acting on the `proxyRes` event
+*  **buffer**: stream of data to send as the request body.  Maybe you have some middleware that consumes the request stream before proxying it on e.g.  If you read the body of a request into a field called 'req.rawbody' you could restream this field in the buffer option:
 
+    ```
+    'use strict';
+
+    const streamify = require('stream-array');
+    const HttpProxy = require('http-proxy');
+    const proxy = new HttpProxy();
+
+    module.exports = (req, res, next) => {
+
+      proxy.web(req, res, {
+        target: 'http://localhost:4003/',
+        buffer: streamify(req.rawBody)
+      }, next);
+
+    };
+    ```
 **NOTE:**
 `options.ws` and `options.ssl` are optional.
 `options.target` and `options.forward` cannot both be missing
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -350,8 +360,41 @@ proxyServer.listen(8015);
        "*": ""
      }
      ```
+*  **cookiePathRewrite**: rewrites path of `set-cookie` headers. Possible values:
+   * `false` (default): disable cookie rewriting
+   * String: new path, for example `cookiePathRewrite: "/newPath/"`. To remove the path, use `cookiePathRewrite: ""`. To set path to root use `cookiePathRewrite: "/"`.
+   * Object: mapping of paths to new paths, use `"*"` to match all paths.
+     For example keep one path unchanged, rewrite one path and remove other paths:
```
donasaur: Minor nit: keep -> to keep

README.md
```diff
@@ -350,8 +360,41 @@ proxyServer.listen(8015);
...
+       "/old.path/": "/new.path/",
+       "*": ""
+     }
+     ```
 *  **headers**: object with extra headers to be added to target requests.
-*  **proxyTimeout**: timeout (in millis) when proxy receives no response from target
+*  **proxyTimeout**: timeout (in millis) for outgoing proxy requests
+*  **timeout**: timeout (in millis) for incoming requests
+*  **followRedirects**: true/false, Default: false - specify whether you want to follow redirects
+*  **selfHandleRequest** true/false, if set to true, none of the webOutgoing passes are called and its your responsibility ro appropriately return the response by listening and acting on the `proxyRes` event
```
donasaur: it's

README.md
```diff
@@ -442,6 +486,33 @@ proxy.close();
 
 ### Miscellaneous
 
+If you want to handle your own response after receiving the proxyRes, you can do
+so with `selfHandleResponse`
```
donasaur: A bit more clarification would be helpful here:
Not sure what it means to handle my own response. Does this mean that I will return my own response to the proxy client? Does the response have to be a modified version of the target webpage response?

A more descriptive var name would also be helpful (e.g., returnCustomResponse).

donasaur: or is selfHandleRequest is the options above the same as selfHandleResponse?

jcrugzz: @donasaur yea good catch it is supposed to just be selfHandleResponse in general. The feature is basically you are taking responsibility for returning whatever you want to the original response after the proxyRequest happens and returns its response. This was the name given by the contributor I'm ok with finding something that makes more sense. Its just trying to indicate that its your responsibility to return the response if thats something you needed to do. `modifyResponse` is also another idea.

jcrugzz:
@Tigge removed that line from the commit. Again, thanks for the contribution!

PR #1264, opened 2018-05-18
[Fix] bodyparser example - support for Content-Type of  application/x-www-form-urlencoded

renxia:
Restream body before proxying, fix to support for Content-Type of  application/x-www-form-urlencoded.
#955 #843 #1027


jcrugzz:
Thanks @renxia

PR #1265, opened 2018-05-20
Fix out sync documentation in http-proxy.js

gaplo917:
I found that the proxy server options readme point to http-proxy.js which the documentation has not been updated.


gaplo917:
no response. just close it

PR #1267, opened 2018-05-23
x-forwarded-host overwrite for mutli level proxies

jaggernoth:
Let's consider a scenario
 `proxy (Internet-facing) -> proxy(load balancer) -> application` where both proxies will be based on node-http-proxy, **with current logic original host to be lost**
With updated logic the original forwarded host will be passed down the proxy chain

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -82,7 +82,7 @@ module.exports = {
         values[header];
     });
 
-    req.headers['x-forwarded-host'] = req.headers['host'] || '';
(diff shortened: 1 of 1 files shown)


jcrugzz:
@jaggernoth host is probably most important but should we also default the other values so we are consistent?

jaggernoth:
@jcrugzz : From what I can see in the code for other x-forwarded header entries new levels are concatenated as CSV values and I was considering making x-forwarded-host CSV as well but it would be much riskier as it may cause compatibility issues (CSV requires additional parsing). 
AFAIK there is no real consensus on how the mechanism should work, but I've done very minimal research just to make my code work;). 
One thing is certain: just overwriting the host seems to defeat the purpose of this header.

PR #1285, opened 2018-07-28
Allow any custom agent instead of just follow-redirects

hasnat:


codebling:
There's already an `agent` option, but it doesn't look like it's fully implemented. What do you think about using `agent` instead of `httpAgent`?

hasnat:
> There's already an `agent` option, but it doesn't look like it's fully implemented. What do you think about using `agent` instead of `httpAgent`?

you're right; agent + httpAgent certainly makes things confusing; possibly docs update to clarify better?


code wise it looks like this
```
http.request(options);
  ^               ^  
  |               |
httpAgent         |
                agent
```
e.g. lib/http-proxy/passes/web-incoming.js:125

my use case was to replace tls with custom module for https requests

codebling:
Oh, I see - this is where we could use something like e.g. `axios` or `got` instead of actually using `http` module's `.request()` with something that `http.Agent`, right? But `httpAgent` here would still need to implement the `http.ClientRequest` interface? So...not `axios` or `got`?  Wouldn't the TLS changes have mostly been specific to the `agent` anyhow? 

There's probably room for refactoring to provide a single configuration that allows both (e.g. hooks for creating connection and getting the stream), but it's probably moot anyhow, since this your PR is going on 4 years old, there have been no commits in almost 2 years. Perhaps a strong ping in the ["New Maintainers" issue](https://github.com/http-party/node-http-proxy/issues/1354) would receive a response, as [indexzero](https://github.com/indexzero) does seem to still be relatively active on GitHub.

hasnat:
thanks, your mentions probably already have alerted maintainers. let's see what's suggested.

PR #1298, opened 2018-10-10
Adds support for protocol relative URLs in location header

bruderstein:
`url.parse` doesn't parse protocol relative URLs, but they are valid
in location headers according to RFC7231
(https://tools.ietf.org/html/rfc7231#section-7.1.2, and specified in
RFC3986 https://tools.ietf.org/html/rfc3986#section-4.2)

This commit fixes that by temporarily adding an `http:` to a URL that
begins with `//` before parsing and removing the protocol before
formatting.

### DIFF
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -51,8 +51,15 @@ module.exports = { // <--
     if ((options.hostRewrite || options.autoRewrite || options.protocolRewrite)
         && proxyRes.headers['location']
         && redirectRegex.test(proxyRes.statusCode)) {
+      var protocolPrefix = '';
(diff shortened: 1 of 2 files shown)


lib/http-proxy/passes/web-outgoing.js
```diff
@@ -68,6 +75,13 @@ module.exports = { // <--
         u.protocol = options.protocolRewrite;
       }
 
+      // If we had to prefix the URL with a protocol,
+      // remove it so we retain the protocol relative
+      // of the response URL
+      if (protocolPrefix) {
+        u.protocol = '';
```
tjenkinson: Maybe it should not be cleared if the 'protocolRewrite' option is used?

odanado:
Are there any updates?

PR #1300, opened 2018-10-22
fix 'Modify Response' readme section to avoid unnecessary array copying

### LINKED ISSUE #1277: "Modify Response" example code has problems
This section of the README: https://github.com/nodejitsu/node-http-proxy#modify-response

It suggests that you should use `body = Buffer.concat([body, data]);` for every chunk received. This causes performance issues for large requests, since the `Buffer.concat` unnecessarily creates a new copy of the Buffer on every chunk.

The NodeJS docs suggest a different approach: pushing each chunk to a mutable array, with only one `Buffer.concat` in the "end" event: https://nodejs.org/en/docs/guides/anatomy-of-an-http-transaction/#request-body

```javascript
let body = [];
request.on('data', (chunk) => {
  body.push(chunk);
}).on('end', () => {
  body = Buffer.concat(body).toString();
});
```

Happy to submit a PR

nilligan:
Fixes #1277

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,12 +501,12 @@ data.
       selfHandleResponse : true
     };
     proxy.on('proxyRes', function (proxyRes, req, res) {
-        var body = new Buffer('');
-        proxyRes.on('data', function (data) {
-            body = Buffer.concat([body, data]);
+        var body = [];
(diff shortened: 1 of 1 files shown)


indexzero:
Thanks!

PR #1307, opened 2018-11-04
remove deprecated 'colors','util' modules

hugetiny:
remove deprecated 'colors','util' modules
replace 'util.puts','util.debug' with console.log and console.error


indexzero:
Thank you for your contribution, but this is doing a few too many changes at once. If you could reopen a PR for just `util.puts` that would be accepted.

PR #1327, opened 2019-03-12
add remove the secure of cookie

buyixiaojiang:
I want to remove 'secure' the property from cookie, so I add the function

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,9 @@ proxyServer.listen(8015);
        "*": ""
      }
      ```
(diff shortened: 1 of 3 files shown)


buyixiaojiang:
I view the wrong info. It is not my code to cause the failed the case. And I revert my code, It's also show the info below.
`
/home/travis/build/nodejitsu/node-http-proxy/node_modules/socket.io/node_modules/debug/src/node.js:132
	let val = process.env[key];
	^^^
`
But I should add the test case for function I wrote. I will add the test case later.

PR #1334, opened 2019-04-16
Fix option.target can't contains question mark

lf541512:
fix case target: `http://www.example.com?project=.example&path=`,and  incoming requrest with query _t=123456


pi0:
This issue has been fixed in [unjs/httpxy#117](https://github.com/unjs/httpxy/pull/117).

PR #1335, opened 2019-05-06
add request source to error handler

simllll:
this allows to see if it's a client error or a backend error. the error handler has next to the target url a new paramter called request source which is either "client" or "target".


simllll:
the failing test is unrelated to this issue I guess....?

simllll:
bump :-)

PR #1395, opened 2019-11-04
FIX: Don't break on invalid headers

asbermudez:
Have found that in some scenarios the proxied server does return invalid characters in the headers, like here:
https://github.com/nodejs/node/issues/21509

In the current version the proxy does break when a invalid character is found with a
`[ERR_INVALID_CHAR]: Invalid character in header content`

This PR aims to avoid the total crash of the application and instead show a warning in the console and process all what can be processed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-  - "6"
-  - "8"
-  - "10"
+  - '8'
+  - '10'
+  - '12'
 script:
   - npm test
 after_success:
--- a/lib/http-proxy/passes/web-outgoing.js
+++ b/lib/http-proxy/passes/web-outgoing.js
@@ -95,7 +95,11 @@ module.exports = { // <--
           if (rewriteCookiePathConfig && key.toLowerCase() === 'set-cookie') {
             header = common.rewriteCookieProperty(header, rewriteCookiePathConfig, 'path');
           }
-          res.setHeader(String(key).trim(), header);
+          try {
+            res.setHeader(String(key).trim(), header);
+          } catch(error) {
+            console.warn(error, key, header);
+          }
         };
 
     if (typeof rewriteCookieDomainConfig === 'string') { //also test for ''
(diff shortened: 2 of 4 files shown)


asbermudez:
The build seems to be breaking because of something I haven't changed... 🤷‍♂

jsmylnycky:
Thanks @asbermudez   This looks reasonable enough to me.  The fail in CI is only during the Node 6 pipeline...but it's past due that Node 6 support is killed off, in my opinion.

asbermudez:
To add some more info, this error specially happens with proxied requests to servers "protected" with [Incapsula](https://www.imperva.com/).
Apparently they do use that invalid character to prevent bots from parsing the page by generating this exception with it.

asbermudez:
@jsmylnycky I've added the changes of https://github.com/http-party/node-http-proxy/pull/1397/files to verify it does build properly.

sshishov:
Hello, when this PR will be merged?

 As we are having the same exact issue and this PR addresses it.

The issue is related to the response from `Incapsula` CDN. We do not want to remove this header as it protects from some bots. We already using `--insecure-http-parser` and `fetch` is working as expected. But unfortunately proxy is not working.

doctornkz:
Hello! Do we have any plans to merge it? We are having a problem with security scanner, it's sending malformed headers and we are getting a storm of exceptions. It's should work more graceful.

dipiash:
Hello! Do you have any plans to merge it? 

We have the same problems

TkDodo:
we've also run into this problem, and we are now applying the fix from this PR via patch-package.

would be good to see this merged sometime, thank you 🙏

conradkirschner:
@jsmylnycky  I was running into this issue and would wish a flag to disable this exception for invalid headers 
(or this got merged somehow )

jsmylnycky:
> @jsmylnycky I was running into this issue and would wish a flag to disable this exception for invalid headers (or this got merged somehow )

Apologies but I'm no longer involved in this project. You'll need to reach out to the current maintainers.

doctornkz-intelas:
RIP

pi0:
This issue has been fixed in [unjs/httpxy#106](https://github.com/unjs/httpxy/pull/106).

PR #1407, opened 2019-12-16
feat onProxyReqWs async handler

LiranBri:
Following issues https://github.com/http-party/node-http-proxy/issues/490 and https://github.com/http-party/node-http-proxy/issues/1403, currently the library does not support async handler for the `proxyReqWs` event for modifying the request before proxying it forward.

This suggested PR solved the problem, without breaking backward compatibility, by introducing a new argument `asyncContext` which is used as follows:

```js 

onProxyReqWs: (proxyReq, req, socket, options, head, asyncContext) =>
      asyncContext(async () => {
        // code ...
        const result = await anyAsyncOperation()
        // code ...
        proxyReq.setHeader('key', result)
      }),

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,8 +513,26 @@ data.
     });
     proxy.web(req, res, option);
 
+```
+
+
+### Modify websockets' proxy request with asynchronous handler
+
+If you want to modify a websockets request with asynchronous hadnler, `proxyReqWs` provides special callback argument named `asyncContext` which gets a single async callback that can return a promise.
+The request would be proxied once the promise is resolved.
+
+```js 
+
+onProxyReqWs: (proxyReq, req, socket, options, head, asyncContext) =>
+      asyncContext(async () => {
+        // code ...
+        const result = await anyAsyncOperation()
+        // code ...
+        proxyReq.setHeader('key', result)
+      }),
 
 ```
+ 
 
 #### ProxyTable API
 
--- a/README.md
+++ b/README.md
@@ -516,7 +516,7 @@ data.
 ```
 
 
-### Modify websockets' proxy request with asynchronous handler
+#### Modify websockets' proxy request with asynchronous handler
 
 If you want to modify a websockets request with asynchronous hadnler, `proxyReqWs` provides special callback argument named `asyncContext` which gets a single async callback that can return a promise.
 The request would be proxied once the promise is resolved.
--- a/lib/http-proxy/passes/ws-incoming.js
+++ b/lib/http-proxy/passes/ws-incoming.js
@@ -105,7 +105,17 @@ module.exports = {
     );
 
     // Enable developers to modify the proxyReq before headers are sent
-    if (server) { server.emit('proxyReqWs', proxyReq, req, socket, options, head); }
+    if (server) {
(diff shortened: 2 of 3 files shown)


LiranBri:
All tests have passed, can we please merge the PR? thanks
@indexzero @jcrugzz @yawnt

asturur:
any development on this PR?

paulrostorp:
Would love to see this merged !

paulrostorp:
I've created a patch for this PR as well as one for the typescript types package:

`http-proxy+1.18.1.patch`:
```js
diff --git a/node_modules/http-proxy/lib/http-proxy/passes/ws-incoming.js b/node_modules/http-proxy/lib/http-proxy/passes/ws-incoming.js
index 270f23f..809ec20 100644
--- a/node_modules/http-proxy/lib/http-proxy/passes/ws-incoming.js
+++ b/node_modules/http-proxy/lib/http-proxy/passes/ws-incoming.js
@@ -76,7 +76,7 @@ module.exports = {
    *
    * @api private
    */
-  stream : function stream(req, socket, options, head, server, clb) {
+  stream : async function stream(req, socket, options, head, server, clb) {
 
     var createHttpHeader = function(line, headers) {
       return Object.keys(headers).reduce(function (head, key) {
@@ -104,9 +104,7 @@ module.exports = {
       common.setupOutgoing(options.ssl || {}, options, req)
     );
 
-    // Enable developers to modify the proxyReq before headers are sent
-    if (server) { server.emit('proxyReqWs', proxyReq, req, socket, options, head); }
-
+    // It is important to register the event listeners of proxyReq before the first await, so we don't have small period of time in which we lose emitted events
     // Error Handler
     proxyReq.on('error', onOutgoingError);
     proxyReq.on('response', function (res) {
@@ -148,7 +146,23 @@ module.exports = {
       server.emit('proxySocket', proxySocket);  //DEPRECATED.
     });
 
-    return proxyReq.end(); // XXX: CHECK IF THIS IS THIS CORRECT
+
+    // Enable developers to modify the proxyReq before headers are sent
+    if (server) {
+      // Provides a way for the event handler to communicate back to the emitter when it finishes its async handling
+      let asyncHandler
+      const asyncContext = (callback) => {
+        asyncHandler = callback
+      }
+
+      server.emit('proxyReqWs', proxyReq, req, socket, options, head, asyncContext);
+
+      if (asyncHandler) {
+        await asyncHandler()
+      }
+    }
+
+    proxyReq.end();
 
     function onOutgoingError(err) {
       if (clb) {
```

`@types+http-proxy+1.17.9.patch`:
```js
diff --git a/node_modules/@types/http-proxy/index.d.ts b/node_modules/@types/http-proxy/index.d.ts
index ca805df..b674ff8 100755
--- a/node_modules/@types/http-proxy/index.d.ts
+++ b/node_modules/@types/http-proxy/index.d.ts
@@ -208,6 +208,7 @@ declare namespace Server {
         socket: net.Socket,
         options: ServerOptions,
         head: any,
+        asyncContext: (cb: Promise) => void
     ) => void;
     type EconnresetCallback<TError = Error, TIncomingMessage = http.IncomingMessage, TServerResponse = http.ServerResponse> = (
         err: TError,
```

PR #1415, opened 2020-01-26
Option to enable fall-through on 404 error

aral:
## Use case: 

  - I’m proxying an app (Hugo)
  - If a URL does not exist there, I want to serve it from an older version of the site

I’ve modified [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) to enable it to fall-through to the middleware chain (see [WIP PR](https://github.com/chimurai/http-proxy-middleware/pull/390), which depends on this one) when it encounters a 404 but without this PR, node-http-proxy already writes out the headers and so we cannot serve the fallback content.

Further context: https://github.com/chimurai/http-proxy-middleware/issues/248#

## Summary of merge request

  - Add `fallthrough` option to the context. If present, 404 responses from the target will not be handled by node-http-proxy so that they can fall through to the middleware chain by the consumer.

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -170,7 +170,7 @@ module.exports = {
     proxyReq.on('response', function(proxyRes) {
       if(server) { server.emit('proxyRes', proxyRes, req, res); }
 
-      if(!res.headersSent && !options.selfHandleResponse) {
+      if(!res.headersSent && !options.selfHandleResponse && (proxyRes.statusCode !== 404)) {
         for(var i=0; i < web_o.length; i++) {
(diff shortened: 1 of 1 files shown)


hpl002:
+1

Any momentum on this PR? Would be great to have this merged

aral:
Just a quick note that I fell down quite a black hole with this and ended up not implementing proxy fallbacks in my project after all. It would be worth weighing the additional complexity against the gains and considering if you can implement what you want to achieve via other methods. Not saying it’s not a legitimate use case, just that it’s not trivial to implement and this PR alone will not achieve it.

hpl002:
That is true. There is a likely a reason for them not implementing it from the start. 

Will see if i can resolve this using other methods.

Thanks for your quick reply

PR #1417, opened 2020-01-31
Allow URL objects as well as strings as options in hostRewrite (http-proxy#1338)

theothermattm:
My first contribution so please let me know where I'm off.  Happy to adjust.


theothermattm:
Build is failing on node6 but I also see an open PR to remove node 6 support, so assuming this is expected?

PR #1419, opened 2020-02-11
adjust web-icoming handling of https protocol checks

alsoscotland:
- inconsistency between regex check and === 'https:' resulted in
    mismatch of secure port and protocol in common.setupOutgoing

https://github.com/http-party/node-http-proxy/issues/1418


pi0:
This issue has been fixed in [unjs/httpxy#123](https://github.com/unjs/httpxy/pull/123).

PR #1433, opened 2020-04-06
fix: check if socket is destroyed before writing to it

### LINKED ISSUE #1432: Uncatchable "This socket has been ended by the other party" when proxying websockets
Sometimes, errors are thrown from line 115 here:

https://github.com/http-party/node-http-proxy/blob/9bbe486c5efcc356fb4d189ef38eee275bbde345/lib/http-proxy/passes/ws-incoming.js#L112-L118

```
Error: This socket has been ended by the other party
    at TLSSocket.writeAfterFIN [as write] (net.js:441:14)
    at ClientRequest.<anonymous> (/home/person/cypress/packages/server/node_modules/http-proxy/lib/http-proxy/passes/ws-incoming.js:115:16)
    at ClientRequest.emit (events.js:210:5)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:583:27)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:115:17)
    at TLSSocket.socketOnData (_http_client.js:456:22)
    at TLSSocket.emit (events.js:210:5)
    at addChunk (_stream_readable.js:308:12)
    at readableAddChunk (_stream_readable.js:289:11)
    at TLSSocket.Readable.push (_stream_readable.js:223:10)
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:182:23)
```

This error can't be caught by passing an error handler to http-proxy.

I think a fix would be to check if the socket is destroyed before attempting to write to it.

Users are able to reproduce this while proxying websockets in real-world apps, it seems to be the result of some race condition as it fails unpredictably: cypress-io/cypress#6458

I haven't been able to come up with a small reproducible example for this bug yet.

flotwig:
Fixes #1432 

`net_socket.destroyed` appears to exist as of at least Node 6.0.0 :+1:

### DIFF
--- a/lib/http-proxy/passes/ws-incoming.js
+++ b/lib/http-proxy/passes/ws-incoming.js
@@ -111,7 +111,7 @@ module.exports = {
     proxyReq.on('error', onOutgoingError);
     proxyReq.on('response', function (res) {
       // if upgrade event isn't going to happen, close the socket
-      if (!res.upgrade) {
+      if (!res.upgrade && !socket.destroyed) {
         socket.write(createHttpHeader('HTTP/' + res.httpVersion + ' ' + res.statusCode + ' ' + res.statusMessage, res.headers));
         res.pipe(socket);
       }


theomessin:
Bump

pi0:
This issue has been fixed in [unjs/httpxy#113](https://github.com/unjs/httpxy/pull/113).

PR #1458, opened 2020-06-16
Fix to use 'abort' event instead of 'aborted' property

### LINKED ISSUE #1455: Error emitted on proxy server after original request is aborted
I found that for Node 10+ if a request is cancelled before being proxied, the proxyServer emits an error instead of an `econnreset` event. This does not happen for < Node 8. I believe this is due to a change in the lifetime of error event listeners on the original request socket.

I think the underlying issue is that in `lib/http-proxy/passes/web-incoming.js` there is an event listener on the `aborted` event on a request. `aborted` is a property, I believe the event should be `abort`. Making this change fixes the issue in my reproduction path.

https://github.com/http-party/node-http-proxy/blob/master/lib/http-proxy/passes/web-incoming.js#L146

Versions:
- Node 12.16.3
- http-proxy 1.17.0
- express 4.17.1

Reproduction path:
- Open server (source code below)
- Open a request to the server (I used cURL, you could use your browser) and immediately stop the request
  - `curl http://127.0.0.1:8080/echo`
- Observe the following error:
```
proxyRequest start
proxyRequest Error: socket hang up
    at connResetException (internal/errors.js:608:14)
    at Socket.socketCloseListener (_http_client.js:400:25)
    at Socket.emit (events.js:322:22)
    at TCP.<anonymous> (net.js:672:12) {
  code: 'ECONNRESET'
}
```

Server code:
```
var httpProxy = require('http-proxy'),
    express = require('express'),
    http = require('http'),
    httpServer,
    app 
        
function sleep(ms) {
  const date = Date.now()
  let currentDate
  while (!currentDate || currentDate - date < ms) currentDate = Date.now()
}       
      
function proxyRequest(request, response) {
  var proxy = httpProxy.createProxyServer({})

  console.log('proxyRequest start')
  
  proxy.on('proxyReq', (proxyReq, req, res, options) => {
    sleep(5000)
  })
  
  proxy.on('error', err => { 
    console.error('proxyRequest', err)
  })
  
  proxy.web(request, response, { target:'127.0.0.1/demo/api/product' })
}   
    
process.title="test"
app = express()
httpServer = http.createServer(app)
  
app.get('/echo', proxyRequest)
    
app.listen('8080') 
```

briannielson:
Fixes https://github.com/http-party/node-http-proxy/issues/1455

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -143,7 +143,7 @@ module.exports = {
     }
 
     // Ensure we abort proxy if request is aborted
-    req.on('aborted', function () {
+    req.on('abort', function () {
       proxyReq.abort();
     });
 
--- a/test/lib-http-proxy-passes-web-incoming-test.js
+++ b/test/lib-http-proxy-passes-web-incoming-test.js
@@ -331,7 +331,7 @@ describe('#createProxyServer.web() using own http server', function () {
       expect(new Date().getTime() - started).to.be.greaterThan(99);
       doneOne();
     });
-    req.end();
+    req.abort();
   });
 
   it('should proxy the request and provide a proxyRes event with the request and response parameters', function(done) {
--- a/test/lib-http-proxy-passes-web-incoming-test.js
+++ b/test/lib-http-proxy-passes-web-incoming-test.js
@@ -331,7 +331,7 @@ describe('#createProxyServer.web() using own http server', function () {
       expect(new Date().getTime() - started).to.be.greaterThan(99);
       doneOne();
     });
-    req.abort();
+    req.end();
   });
 
   it('should proxy the request and provide a proxyRes event with the request and response parameters', function(done) {


briannielson:
Closing due to failing test.

briannielson:
Tests fail, but they need to be reworked to use a ClientRequest object as specified in `web-incoming.js`:
```
  /**
   * Does the actual proxying. If `forward` is enabled fires up
   * a ForwardStream, same happens for ProxyStream. The request
   * just dies otherwise.
   *
   * @param {ClientRequest} Req Request object
   * @param {IncomingMessage} Res Response object
   * @param {Object} Options Config object passed to the proxy
   *
   * @api private
   */
```

jayeshmend:
@briannielson any idea when this will be fixed as its creating issues at random interval.

briannielson:
> @briannielson any idea when this will be fixed as its creating issues at random interval.

Sorry, I haven't been watching this PR or the related issue. We ended up patching the library in our codebase with this change. This caused tests to fail in this repo, but I didn't have time to spend on fixing the tests. This did appear to fix the issue in our production environment.

Issue #1451, opened 2020-06-01
More intelligent selfHandleResponse 


lukenofurther:
In my use case I need to do some custom transformations on the response but only for a small subset of requests. 

I can perform custom processing on the response if I set `selfHandleResponse` to true. However, for all requests that I don't want to do anything custom to, I must copy this library's source code and do [the hacky workaround suggested on this other issue](https://github.com/http-party/node-http-proxy/issues/1263#issuecomment-394758768).

That hacky solution works at the moment but is bad, easily broken by a future update to the library.

I suggest allowing a more customisable value for the `selfHandleResponse` option. The simplest option would be accepting a function (in addition to true/false), which when supplied is used to check whether to perform the default incoming processing or leave it to be handled in a custom manner.

Is there an alternative non-hacky way for me to achieve what I need to achieve? Would the above be a valuable enhancement?

ctrlplusb:
Wow, 18 hours ago. Perfect timing. I am needing something very similar. 😅

Pines-Cheng:
I meet the same problem, I just want to perform custom processing on the part of the response, and return the default response in the rest of response, how can I get that?

PR #1461, opened 2020-06-30
feat: add forcePasses and selfHandleResponse as function callback

### LINKED ISSUE #1263: selfHandleResponse: true => response broken
I'm trying to manually handle the response with `{ selfHandleResponse: true }`, but, as the doc says:
> none of the webOutgoing passes are called

And that prevents me from correctly sending response.

This is the code causing the problem:

```js
// passes/web-incoming.js
if(!res.headersSent && !options.selfHandleResponse) {
  for(var i=0; i < web_o.length; i++) {
    if(web_o[i](req, res, proxyRes, options)) { break; }
  }
}
```
Doing this, fixes it:

```js
if(!res.headersSent) {
  for(var i=0; i < web_o.length; i++) {
    if(web_o[i](req, res, proxyRes, options)) { break; }
  }
}
```

I need to call the webOutgoing passes and do some extra stuff. But I can't.
Maybe you can expose `passes/web-outgoing`, or add a flag like `callWebOutgoing` to let users call them if they want to.

phawxby:
Fixes #1263 & #1451.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,7 @@ proxyServer.listen(8015);
 *  **timeout**: timeout (in millis) for incoming requests
 *  **followRedirects**: true/false, Default: false - specify whether you want to follow redirects
 *  **selfHandleResponse** true/false, if set to true, none of the webOutgoing passes are called and it's your responsibility to appropriately return the response by listening and acting on the `proxyRes` event
+*  **forcePasses** true/false, if set to true the web passes will be run even if `selfHandleResponse` is also set to true.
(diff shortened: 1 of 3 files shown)


phawxby:
@jcrugzz @indexzero I don't suppose there's any chance of getting this merged in and released fairly quickly?

jennydaman (approved):

JounQin:
I've published it as `http-proxy-x` at https://github.com/un-es/http-proxy-x.

PR #1494, opened 2020-12-21
feat: when get a chunked response, flush headers

Yohox:
if proxy a chunked response. when server return http-header but don't response body immediately. http-proxy will blocked until server body sent.


JoaoOtavioS:
Hi, I took the liberty of implementing this in my fork. https://github.com/squarecloudofc/http-proxy

PR #1495, opened 2020-12-23
do not pipe req into proxyReq until socket is connected

pepmartinez:
While building a reverse proxy with ha-lb capabilities I found out that proxy.web() does not play totally well with retries if a request body is present: it turns out the original req is piped into the proxyReq immediately, even before the socket at proxyReq tries to connect; such piping would read some (or even all) of the original req's body even if the socket does not connect; those body bytes would be lost if we choose to proxy.web() the same request onto an alternative target

The patch delays the piping until proxyReq's socket is present, and it's either connected (that is, reused from an agent) or suceeds to connect

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -129,6 +129,15 @@ module.exports = {
 
     // Enable developers to modify the proxyReq before headers are sent
     proxyReq.on('socket', function(socket) {
+      if (socket.pending) {
+        // if not connected, wait till connect to pipe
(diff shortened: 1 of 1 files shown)


chasetec:
Don't you need to emit the proxyReq event before the pipe call? If the pipe sends the first chunk then the headers will be sent with it and won't be mutable when the proxyReq event is generated.

pepmartinez:
Well, all I did is to defer the pipe() until there's a socket. There were no emit() before, to start with

pi0:
This issue has been fixed in [unjs/httpxy#111](https://github.com/unjs/httpxy/pull/111).

PR #1504, opened 2021-02-02
fix(outgoing-opts): preference for values in ssl

### LINKED ISSUE #1503: Outgoing SSL options overwritten by undefined values from main options
https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/common.js#L37-L40

stanleynguyen:
Preventing overwriting by undefined first-layer key from the main options
Resolves #1503


pi0:
This issue has been fixed in [unjs/httpxy#118](https://github.com/unjs/httpxy/pull/118).

PR #1513, opened 2021-04-14
setRedirectHostRewrite options.target support URL Object

ksky521:
in https://github.com/http-party/node-http-proxy/blob/master/lib/http-proxy/index.js#L64-L65 `options.target` can use `URL` instance: 
```js
const url = new URL(req.url);
proxyServer.web(req, res, {
    target: url
});
```
https://github.com/http-party/node-http-proxy/blob/master/lib/http-proxy/passes/web-outgoing.js#L54 don't support URL instance


pi0:
Hi. This is possible with [unjs/httpxy](https://github.com/unjs/httpxy) fork to pass a URL.

PR #1542, opened 2021-07-21
Fix handling of client disconnect so that econnreset event is emitted instead of error

### LINKED ISSUE #1455: Error emitted on proxy server after original request is aborted

gsf4726:
Fixes #1455

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -154,7 +154,7 @@ module.exports = {
 
     function createErrorHandler(proxyReq, url) {
       return function proxyError(err) {
-        if (req.socket.destroyed && err.code === 'ECONNRESET') {
+        if (req.aborted && err.code === 'ECONNRESET') {
           server.emit('econnreset', err, req, res, url);
           return proxyReq.abort();
         }
--- a/test/lib-http-proxy-passes-web-incoming-test.js
+++ b/test/lib-http-proxy-passes-web-incoming-test.js
@@ -334,6 +334,56 @@ describe('#createProxyServer.web() using own http server', function () {
     req.end();
   });
 
+  it('should proxy the request and handle client disconnect error', function(done) {
+    var proxy = httpProxy.createProxyServer({
+      target: 'http://127.0.0.1:45002',
+    });
+
+    require('net').createServer().listen(45002);
+
+    var proxyServer = http.createServer(requestHandler);
+
+    var cnt = 0;
+    var doneOne = function() {
+      cnt += 1;
+      if(cnt === 2) done();
+    }
+
+    var started = new Date().getTime();
+    function requestHandler(req, res) {
+      proxy.once('econnreset', function (err, errReq, errRes) {
+        proxyServer.close();
+        expect(err).to.be.an(Error);
+        expect(errReq).to.be.equal(req);
+        expect(errRes).to.be.equal(res);
+        expect(err.code).to.be('ECONNRESET');
+        doneOne();
+      });
+
+      proxy.web(req, res);
+    }
+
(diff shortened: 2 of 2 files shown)


edspencer:
Thank you so much for this fix; this issue has been the bane of my life for weeks and your solution totally fixes it. Would love to see this merged ASAP so I don't have to keep on modifying my local http-proxy files.

franckc:
Fix works great. I have been using it on my project. Would be awesome to see this merged in! Please.... :)

edspencer:
Looks like this project is dead, so in order to use this fix we just forked the project and made the 1 line change in this PR. It seems to work for us. Took about 30 mins, but better than manually editing that file in node_modules 🤦

ReyRod (approved):

ReyRod:
As this project is dead, we had to apply this change with https://www.npmjs.com/package/patch-package, thanks for the solution @gsf4726

sidx1024:
Monkey patching function for who don't want to use the patch-package.

```js
function monkeyPatchHttpProxyLibrary() {
  // https://github.com/http-party/node-http-proxy/pull/1542
  const filePath = path.join(
    __dirname,
    '../node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js',
  );
  const fileContent = fs.readFileSync(filePath, 'utf-8');
  const patchedFileContent = fileContent.replace(
    `if (req.socket.destroyed && err.code === 'ECONNRESET') {`,
    `if (req.aborted && err.code === 'ECONNRESET') {`,
  );
  fs.writeFileSync(filePath, patchedFileContent, 'utf-8');
}
```

pi0:
This issue has been fixed in [unjs/httpxy#115](https://github.com/unjs/httpxy/pull/115).

PR #1551, opened 2021-09-10
header could be empty string

damiandragowski:
changing condition, do setHeader only when header is not 

null
undefined
NaN
empty string ("")
0
false

It preserve from exception:

[HPM] GET /events?uuids=18cd56cb-45cb-43e6-a397-c4531cc57866 -> http://localhost:8099
_http_outgoing.js:545
    throw new ERR_INVALID_HTTP_TOKEN('Header name', name);
    ^

TypeError [ERR_INVALID_HTTP_TOKEN]: Header name must be a valid HTTP token [""]
    at ServerResponse.setHeader (_http_outgoing.js:563:3)


pi0:
This issue has been fixed in [unjs/httpxy#121](https://github.com/unjs/httpxy/pull/121).

PR #1552, opened 2021-09-22
Fix websocket socket close before response

FinnStutzenstein:
There are situations where the socket can already be closed. Writing to
it will result in a "This socket has been ended by the other party"
error which results in the node service to terminate (cannot be catched
by a try-catch block).


ravin00 (approved):

pi0:
This issue has been fixed in [unjs/httpxy#114](https://github.com/unjs/httpxy/pull/114).

PR #1559, opened 2021-12-08
Fix 'aborted' detection on Node v15.5.0+

Jimbly:
This fixes a major memory leak I encountered usign this on Node v16.
Doing a binary search of node versions, the problem appears to originate in v15.5.0 as a result of nodejs/node#33035, however later changes have quietly completely removed the 'aborted' event that `http-proxy` relies on, and later added a deprecation note about it (which seems to actually be incorrect).
Despite what the notes about [DEP0156](https://nodejs.org/dist/latest-v16.x/docs/api/deprecations.html#DEP0156) say, the only way I could get this module working reliably again was to replace `req.on('aborted')` with instead checking `res.on('close')` and looking at `res.writeableFinished`.

For a very easy test, load any video file over the proxy, and seek backwards in Chrome.  For example, run the following:
```javascript
const httpProxy = require('http-proxy');

httpProxy.createProxyServer({
  target: 'http://files.dashingstrike.com.s3.amazonaws.com',
  changeOrigin: true,
}).listen(8080);
```

And then open http://localhost:8080/SplodyGameplayLong.mp4, and then seek backwards in the file.  Chrome will abort hundreds of requests (each of which is requesting ~32MB), which, if not properly aborted, will leak connections as well as hundreds of MBs of TCP kernel memory per second (quickly bricking your instance, making it unreachable even by SSH, as I found out -_-).  On Linux, you can view TCP kernel memory used with `cat /proc/net/sockstat` (`mem` is count of 4K pages), or `ss -atn state big` if you have `ss` installed.

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -18,6 +18,11 @@ var nativeAgents = { http: httpNative, https: httpsNative };
  * flexible.
  */
 
+// 'aborted' event stopped working reliably on v15.5.0 and was later removed entirely
+var supportsAbortedEvent = (function () {
+  var ver = process.versions.node.split('.').map(Number);
+  return ver[0] <= 14 || ver[0] === 15 && ver[1] <= 4;
+}());
 
 module.exports = {
 
@@ -143,9 +148,18 @@ module.exports = {
     }
 
     // Ensure we abort proxy if request is aborted
-    req.on('aborted', function () {
-      proxyReq.abort();
-    });
+    if (supportsAbortedEvent) {
+      req.on('aborted', function () {
+        proxyReq.abort();
+      });
+    } else {
+      res.on('close', function () {
+        var aborted = !res.writableFinished;
+        if (aborted) {
+          proxyReq.abort();
+        }
+      });
+    }
 
     // handle errors in proxy and incoming request, just like for forward proxy
     var proxyError = createErrorHandler(proxyReq, options.target);


RemiBou:
Hi @Jimbly any plan to merge this ?

Jimbly:
> Hi @Jimbly any plan to merge this ?

I'm not the one who can merge to this repository, that would be a question for the maintainers of `node-http-proxy`.

Since there was no response here (abandoned project?) I had to publish my fork for some other projects though, and you can use that in the meantime: [`http-proxy-node16`](https://www.npmjs.com/package/http-proxy-node16)

RemiBou:
Thanks I'll checkit out ! Great work :)

Le mar. 5 juil. 2022 à 14:59, Jimb Esser ***@***.***> a
écrit :

craftzdog:
This patch worked! Thanks for the fix

likev:
The response 'close' event and the 'destroy' method are the correct alternatives
https://github.com/http-party/node-http-proxy/pull/1580

anthonyalayo:
Is the maintainer gone?

Jimbly:
No maintainers as far as I can tell.  I've (just now) pushed an update to [http-proxy-node16](https://www.npmjs.com/package/http-proxy-node16) so that it correctly links back to [my fork](https://github.com/Jimbly/http-proxy-node16) instead of this repo, if anyone wants to redirect PRs to that repository, I can potentially merge them and update my fork (with the caveat that I don't really know any of the internals here ^_^).

cassepipe:
Does this affect all versions above 16 or just 16 ?

Is this why I have a 
```json
{"statusCode":403,"message":"sse already opened","error":"Forbidden"}
```
... when using vite ?

Jimbly:
This affects all versions 16 and above, it's due to a low-level, non-backwards-compatible change in Node.  No idea if it would have anything to do with that SSE error, though you can try swapping out for `http-proxy-node16` and see if it fixes it, it should be otherwise the same package.

cassepipe:
This seems like a no-brainer to merge this. Is there any issues with it @indexzero @jsmylnycky  ?

You could either merge this one which is apparently backwards compatible with older versions of `node` or this drop backwards compatibility with this shorter one https://github.com/http-party/node-http-proxy/pull/1580.

Then you can kill two merge request with one stone and close whichever you don't merge.

lib/http-proxy/passes/web-incoming.js
```diff
@@ -18,6 +18,11 @@ var nativeAgents = { http: httpNative, https: httpsNative };
  * flexible.
  */
 
+// 'aborted' event stopped working reliably on v15.5.0 and was later removed entirely
+var hasAborted = (function () {
```
callumgare: Super minor suggestion but when I first glanced at this based on the name I presumed it was a function for checking if a connection had been aborted. So maybe something like "supportsAbortedEvent" might be a bit of a clearer name?

Jimbly: Sounds good, changed!

pickeloe5:
My website was crashing while I was advertising it which made me very upset.

This P.R. fixes the crash, thank you very much Jimbly.

lib/http-proxy/passes/web-incoming.js
```diff
@@ -18,6 +18,11 @@ var nativeAgents = { http: httpNative, https: httpsNative };
  * flexible.
  */
 
+// 'aborted' event stopped working reliably on v15.5.0 and was later removed entirely
```
ravin00, line 21: `var supportsAbortedEvent = (function () {
  var [major, minor] = process.versions.node.split('.').map(Number);
  return major < 15 || (major === 15 && minor < 5);
}());`

i think this will be more suitable

lib/http-proxy/passes/web-incoming.js
```diff
@@ -143,9 +148,18 @@ module.exports = {
     }
 
     // Ensure we abort proxy if request is aborted
-    req.on('aborted', function () {
-      proxyReq.abort();
-    });
+    if (supportsAbortedEvent) {
```
ravin00, line 151: `if (supportsAbortedEvent) { req.on('aborted', () => proxyReq.abort()); } else { res.on('close', () => { if (!res.writableFinished) proxyReq.abort(); }); }
`
This would be suitable too i think

ravin00 (changes requested):

ImranR-TI:
>  Chrome will abort hundreds of requests

Do you know how to reproduce this programatically? I tried "aborting" a long running request using `timeout 1s curl ...` and `curl --connect-timeout 1 --max-time 1 ...` but those seem to fail correctly -- no leaked sockets (the Chrome method does work for me).

Jimbly:
I'm not sure what difference there would be between killing the socket via process exiting like with curl and other forms of aborting (is there an HTTP2 abort that Chrome uses maybe, keeping the socket open?), but you could try a node script doing `fetch()` and [`AbortController::abort()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) (and maybe not immediately exiting the process, as that may cause the OS to clean up the socket in a more heavy-handed way)...

teklakct:
Hi @eriktrom do you have time to check this? Or maybe you can consider adding some additional maintainers

teklakct:
> Hi @eriktrom do you have time to check this? Or maybe you can consider adding some additional maintainers

or maybe @BigBlueHat @indexzero @thornjad

eriktrom:
> Hi @eriktrom do you have time to check this? Or maybe you can consider adding some additional maintainers

I don't have more permissions than the rest of us for this repo. Sorry.

teklakct:
> > Hi @eriktrom do you have time to check this? Or maybe you can consider adding some additional maintainers
> 
> I don't have more permissions than the rest of us for this repo. Sorry.

Thanks for the quick reply. I saw your activity on another repository and assumed you had it. Then the forked one is still an option

pi0:
This issue has been fixed in [unjs/httpxy#110](https://github.com/unjs/httpxy/pull/110).

PR #1580, opened 2022-04-06
fix request abort for node 16+

Matt-Esch:
After upgrading to node 16, the web proxy stopped working when requests were aborted. The specific use case was SSE where the incoming request has ended but the server response is a long lived http connection. The 'aborted' event and the 'abort' method are deprecated. The response 'close' event and the 'destroy' method are the correct alternatives and pre-date node 8, which is the minimum version this package currently supports. I did run tests on node 8/10/12/14/16 and they failed on node 16 prior to this fix and all pass after this fix.

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -143,8 +143,8 @@ module.exports = {
     }
 
     // Ensure we abort proxy if request is aborted
-    req.on('aborted', function () {
-      proxyReq.abort();
+    res.on('close', function () {
+      proxyReq.destroy();
     });
 
     // handle errors in proxy and incoming request, just like for forward proxy


kulek1 (approved):

anthonyalayo:
Can we merge this?

callumgare:
Unfortunately won't work in node.js 16 and below since the "close" event was added to in v16:

https://nodejs.org/docs/latest-v16.x/api/http.html#event-close
https://nodejs.org/docs/latest-v15.x/api/http.html#event-close

Can't find any info on what's the minimum node version node-http-proxy supports but in this case it should be pretty easy to use "abort" for v15 and below and "close" for v16 and above. That's what https://github.com/http-party/node-http-proxy/pull/1559 does

gdrbyKo1:
@callumgare I'm not sure I follow. This PR attaches a "close" event listener to `res`, which is an instance of `http.ServerResponse`. According to Node.js docs, this event is present in both v16 and v15:
https://nodejs.org/docs/latest-v16.x/api/http.html#event-close_2
https://nodejs.org/docs/latest-v15.x/api/http.html#http_event_close_1
Am I missing something?

pi0:
This issue has been fixed in [unjs/httpxy#107](https://github.com/unjs/httpxy/pull/107).

PR #1581, opened 2022-04-20
[fix] do not append XHeader values

yulodl:


pi0:
This issue has been fixed in [unjs/httpxy#120](https://github.com/unjs/httpxy/pull/120).

PR #1600, opened 2022-07-16
crash when 'options.target' is an object instead a string

vilherda:
Sometimes the attribute 'options.target' is not a string and is an object like:
```json
{
    "protocol": "http:",
    "slashes": true,
    "auth": null,
    "host": "some.hostname.fake",
    "port": null,
    "hostname": "some.hostname.fake",
    "hash": null,
    "search": null,
    "query": null,
    "pathname": "/",
    "path": "/",
    "href": "http://some.hostname.fake"
}
```
In this case is necessary to access to the attribute 'options.target.href'.


Cuadrix (approved):

pi0:
This issue has been fixed in [unjs/httpxy#108](https://github.com/unjs/httpxy/pull/108).

PR #1608, opened 2022-09-01
Updated Version of follow-redirects from 1.0.0 to 1.14.8

Annniee1:
We are facing bugs in older version .Please Upgrade the version

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "eventemitter3": "^4.0.0",
     "requires-port": "^1.0.0",
-    "follow-redirects": "^1.0.0"
+    "follow-redirects": "^1.14.8"
   },
   "devDependencies": {
     "async": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
   "main": "index.js",
   "dependencies": {
     "eventemitter3": "^4.0.0",
-    "requires-port": "^1.0.0",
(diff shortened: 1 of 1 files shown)


Annniee1:
@peters @darashi @benatkin @indexzero Please Review this Pull Request

Annniee1:
Hi @peters @darashi @benatkin @indexzero @Jimbly. Any Plan to merge this Pull Request?

Jimbly:
I'm not a maintainer of this repo (as far as I know, it's dead), however I published [http-proxy-node16](https://www.npmjs.com/package/http-proxy-node16) a while back, which is a fork of this repo that works on Node 16 without gigantic memory leaks (#1559).  Feel free to send a PR [over there](https://github.com/Jimbly/node-http-proxy) and I'll publish it to NPM.

Jimbly:
That being said, I believe `follow-redirects v1.14.8` matches the semver specified here already, so I'm not sure this change is needed - if you install v1.14.8, it should use it, I think?

Annniee1:
> That being said, I believe `follow-redirects v1.14.8` matches the semver specified here already, so I'm not sure this change is needed - if you install v1.14.8, it should use it, I think?

Yaa @Jimbly . It matches the semver version specified. I am not sure why previously i was getting 1.14.7 and not 1.14.8.
But yes, it is working now.
Thanks!

PR #1629, opened 2023-03-07
add a showcase how to proxy POST with a body

capaj:
props to @Tokscull

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,42 @@ http.createServer(function (req, res) {
 
 **[Back to top](#table-of-contents)**
 
+#### Forward POST requests with a json body
+
+```ts
+const proxy = httpProxy.createProxyServer({
+  target: `http://localhost:3002`
+})
+
+proxy.on('proxyReq', (proxyReq, req, res, options) => {
+  // @ts-expect-error
+  if (req.body) {
+    // @ts-expect-error
+    let bodyData = req.rawBody
+    proxyReq.setHeader('Content-Type', 'application/json')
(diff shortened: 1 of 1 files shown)


alanhe421:
so any update

README.md
```diff
@@ -228,6 +228,42 @@ http.createServer(function (req, res) {
...
+```ts
+const proxy = httpProxy.createProxyServer({
+  target: `http://localhost:3002`
+})
+
+proxy.on('proxyReq', (proxyReq, req, res, options) => {
+  // @ts-expect-error
+  if (req.body) {
+    // @ts-expect-error
+    let bodyData = req.rawBody
```
ravin00, line 242: unsafe access to req.rawBody

README.md
```diff
@@ -228,6 +228,42 @@ http.createServer(function (req, res) {
 
 **[Back to top](#table-of-contents)**
 
+#### Forward POST requests with a json body
+
+```ts
```
ravin00, line 233: `import { IncomingMessage, ServerResponse } from 'http';
import * as httpProxy from 'http-proxy';`

add this to the first

README.md
```diff
@@ -228,6 +228,42 @@ http.createServer(function (req, res) {
...
+const proxy = httpProxy.createProxyServer({
```
ravin00, line 234: Before the `const proxy = httpProxy.createProxyServer` create an interface `interface ExtendedIncomingMessage extends IncomingMessage {
  body?: any;
  rawBody?: Buffer | string;
}`

the reason to create this interface is to typeScript will catch errors if you try to access properties that don't exist. and to make it runtime error free

PR #1650, opened 2023-10-07
feat: allow sending a custom proxy timeout error

### LINKED ISSUE #1331: Specific error for timeouts
I'm currently proxying requests like this: 

```
this._proxy.web(req, res, { target: target.host, proxyTimeout }, err => {
				logger.error("Error proxying %s -> { Host: %s, URL: %s }: ", req.originalUrl, target.host, target.requestURL);
				logger.error(err);
				res.status(500).send();
			});
```

If the upstream server takes too long, the callback is run with the following error:

```
2019-03-28T19:49:22.165Z - error:  message=socket hang up, stack=Error: socket hang up
    at createHangUpError (_http_client.js:331:15)
    at Socket.socketCloseListener (_http_client.js:363:23)
    at emitOne (events.js:121:20)
    at Socket.emit (events.js:211:7)
    at TCP._handle.close [as _onclose] (net.js:557:12)
, code=ECONNRESET
```

Would it make sense to get a timeout specific error? so I can, if I chose to do so, return a specific timeout error like 504?

rowanmanning:
When using `proxyTimeout` it's very difficult to tell the difference between a regular socket hangup and a timeout because, in both cases, an `ECONNRESET` error is thrown. Ideally we should be able to identify when a proxy request has failed because it took too long, this would allow us to do things like send appropriate `504` status code.

Suddenly throwing a different error would probably be considered a breaking change because it's possible that users of http-proxy are relying on the `ECONNRESET` error. I decided to add the custom timeout error behind a new option for now so that people can opt into using it.

If you set this option:

```js
var proxy = httpProxy.createProxyServer({
  target: 'http://example.com',
  proxyTimeout: 100,
  proxyTimeoutCustomError: true
});
```

Then the error that gets thrown will have a message of `"The proxy request timed out"` and a code of `ETIMEDOUT` to match Node.js: https://nodejs.org/api/errors.html#common-system-errors

This allows for custom error handling code like this:

```js
proxy.on('error', function(err, req, res) {
  if (err.code === 'ETIMEDOUT') {
    res.writeHead(504);
  } else {
    res.writeHead(503);
  }
  // ...
});
```

## A note on implementation

I had to switch from [`request.abort()`](https://nodejs.org/api/http.html#requestabort) to [`request.destroy()`](https://nodejs.org/api/http.html#requestdestroyerror) for this to work. `request.abort()` has been deprecated since Node.js v14.1.0 and `request.destroy()` has been available since Node.js v0.3.0 so I think this is a safe bet. If you're worried about it then I could always switch to `request.abort()` for the case when `proxyTimeoutCustomError` is falsy.

---

Resolves #1331.

### DIFF
--- a/lib/http-proxy/passes/web-incoming.js
+++ b/lib/http-proxy/passes/web-incoming.js
@@ -138,7 +138,12 @@ module.exports = {
     // show an error page at the initial request
     if(options.proxyTimeout) {
       proxyReq.setTimeout(options.proxyTimeout, function() {
-         proxyReq.abort();
+        if (options.proxyTimeoutCustomError) {
+          var timeoutError = new Error('The proxy request timed out');
+          timeoutError.code = 'ETIMEDOUT';
+          return proxyReq.destroy(timeoutError);
+        }
+        proxyReq.destroy();
       });
     }
 
--- a/test/lib-http-proxy-passes-web-incoming-test.js
+++ b/test/lib-http-proxy-passes-web-incoming-test.js
@@ -287,6 +287,41 @@ describe('#createProxyServer.web() using own http server', function () {
     }, function() {}).end();
   });
 
+  it('should proxy the request with custom timeout errors (proxyTimeoutCustomError)', function(done) {
+    var proxy = httpProxy.createProxyServer({
+      target: 'http://127.0.0.1:45002',
+      proxyTimeout: 100,
+      proxyTimeoutCustomError: true
+    });
+
+    require('net').createServer().listen(45002);
+
+    var proxyServer = http.createServer(requestHandler);
+
+    var started = new Date().getTime();
+    function requestHandler(req, res) {
+      proxy.once('error', function (err, errReq, errRes) {
+        proxyServer.close();
+        expect(err).to.be.an(Error);
+        expect(errReq).to.be.equal(req);
+        expect(errRes).to.be.equal(res);
+        expect(new Date().getTime() - started).to.be.greaterThan(99);
+        expect(err.code).to.be('ETIMEDOUT');
+        done();
+      });
+
+      proxy.web(req, res);
+    }
+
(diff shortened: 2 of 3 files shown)


JoaoOtavioS:
Hi, I took the liberty of implementing this in my fork. https://github.com/squarecloudofc/http-proxy

lib/http-proxy/passes/web-incoming.js
```diff
@@ -138,7 +138,12 @@ module.exports = {
     // show an error page at the initial request
     if(options.proxyTimeout) {
       proxyReq.setTimeout(options.proxyTimeout, function() {
-         proxyReq.abort();
+        if (options.proxyTimeoutCustomError) {
+          var timeoutError = new Error('The proxy request timed out');
+          timeoutError.code = 'ETIMEDOUT';
+          return proxyReq.destroy(timeoutError);
+        }
+        proxyReq.destroy();
```
ravin00, line 146: what's the change here?

test/lib-http-proxy-passes-web-incoming-test.js
```diff
@@ -287,6 +287,41 @@ describe('#createProxyServer.web() using own http server', function () {
...
+    var proxyServer = http.createServer(requestHandler);
+
+    var started = new Date().getTime();
+    function requestHandler(req, res) {
+      proxy.once('error', function (err, errReq, errRes) {
+        proxyServer.close();
+        expect(err).to.be.an(Error);
+        expect(errReq).to.be.equal(req);
+        expect(errRes).to.be.equal(res);
+        expect(new Date().getTime() - started).to.be.greaterThan(99);
```
ravin00, line 308: what about adding the `expect(new Date().getTime() - started).to.be.above(99);
`. Will it make a difference to this line or to the whole code

ravin00 (changes requested):

PR #1668, opened 2024-05-09
fix: Avoid using deprecated util._extend()

danjoa:
Node 22 dumps a deprecation warning when using `util._extend()`:

```
[DEP0060] DeprecationWarning: The `util._extend` API is deprecated. 
Please use Object.assign() instead.
```


danjoa:
- Duplicate to https://github.com/http-party/node-http-proxy/pull/1666

PR #1692, opened 2024-11-23
fix: replace deprecated util._extend with Object.assign()

Auios:
The functionality remains identical as both methods perform shallow copies of object properties. `Object.assign()` is the standard modern approach for this operation.

Resolves this warning:
```
[DEP0060] DeprecationWarning: The `util._extend` API is deprecated. Please use Object.assign() instead.
```


Auios:
Noticed this was a duplicate of https://github.com/http-party/node-http-proxy/pull/1666

PR #1699, opened 2025-01-11
fix: deprecation DEP0060, use Object.extend instead of util._extend

kwikwag:
See https://nodejs.org/api/deprecations.html#DEP0060


kwikwag:
Hmm, apparently dup of https://github.com/http-party/node-http-proxy/pull/1666

